PR #2, opened 2018-05-06
feat(TextMatch): upgrade dom-testing-library

kentcdodds:
- Changes queries to default to exact string matching
- Can opt-in to fuzzy matches by passing { exact: true } as the last arg
- Queries that search inner text collapse whitespace
  (queryByText, queryByLabelText)

BREAKING CHANGE: Strings are considered to be an exact match now. You can opt-into fuzzy matching, but it's recommended to use a regex instead.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "license": "MIT",
   "dependencies": {
     "@vue/test-utils": "^1.0.0-beta.15",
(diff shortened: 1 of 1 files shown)


dfcook:
I'd manually done this change before I saw this PR, sorry.

PR #3, opened 2018-10-24
Fix vue-router link in readme

lindgr3n:
Fixes `vue-router` link in readme to point to correct path.

Just starting to try it out a bit. Easy to get going, great work! 👍


dfcook:
Thanks :)

PR #7, opened 2018-11-26
fix routes not rendering

ankitsinghaniyaz:
Fixes https://github.com/dfcook/vue-testing-library/issues/6

Issue:
- routes should be passes as an object

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,10 +1,21 @@
-import { createLocalVue, mount } from '@vue/test-utils'
+import {
+  createLocalVue,
+  mount
+} from '@vue/test-utils'
 import Simulate from './Simulate'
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -18,7 +29,9 @@ function render (TestComponent, { props = null, store = null, routes = null } =
   if (routes) {
     const VueRouter = require('vue-router')
     localVue.use(VueRouter)
-    router = new VueRouter(routes)
+    router = new VueRouter({
```
ankitsinghaniyaz: This is the change. Had to pass routes as an object.

As outlined in this [example](https://lmiller1990.github.io/vue-testing-handbook/vue-router.html#writing-the-test)

PR #12, opened 2018-12-27
Enables child component stubs

tim-maguire:
Hello Daniel. 

This PR enables the [vue-test-utils `stubs` options](https://vue-test-utils.vuejs.org/api/options.html#stubs). 

Example use case

This: 
```js
import { FontAwesomeIcon } from "@fortawesome/vue-fontawesome"
import { library, config } from "@fortawesome/fontawesome-svg-core"
import { faSearch } from "@fortawesome/free-solid-svg-icons"
import { render, cleanup } from "../../src"
import Form from "./components/Form"

afterEach(cleanup)

library.add(faSearch)

test("Form contians with search button", () => {
  const { getByText } = render(Form, {}, vue => {
    vue.component(FontAwesomeIcon.name, FontAwesomeIcon)
  })
  getByText("Search now")
})
```

Can reduce to:
```js
import { render, cleanup } from "../../src"
import Form from "./components/Form"

afterEach(cleanup)

test("Form contians with search button", () => {
  const { getByText } = render(Form, {
    stubs: ['font-awesome-icon']
  })
  getByText("Search now")
})
```

PR #13, opened 2018-12-27
Support all mount options in vue-test-utils

dfcook:
Allow all supported options to be passed to vue-test-utils mount function. Replaces #12 , hence including test from that PR in here.


dfcook:
There is, should we bite the bullet and do it here making it a breaking change?

dfcook:
Pushed a further commit that will allow props or propsData to be passed and will copy props to propsData, this will keep backwards compatibility and allow propsData to be used going forward.

tim-maguire:
Yeah, this seems like a sensible approach.

PR #26, opened 2019-05-28
Rename index file to simplify test imports

afontcu:
Hi!

By renaming the sole file of vue-testing-library, and making use of jest' moduleDirectories, example tests might look a bit simpler. If this PR is merged, tests would be "copy-pastable" into another project.

btw, I assumed that modifying the `main` key in `package.json` would make everything work as expected. If that's not the case and you are still interested in this PR, point me to the right direction and I'll change it.

Thanks! 😃


dfcook:
Thanks for your recent PRs, sorry it's taken a while to get around to looking at them.

afontcu:
@dfcook oh no, thank you! Really happy to help. Feel free to ping me if you need some help or if you know any are that could use some love!

PR #29, opened 2019-05-31
Feature/scoped packages

dfcook:
Move the library to use scoped packages


afontcu:
Given the name change, should `src/vue-testing-library.js` be renamed and (test files updated accordingly) as in #26? The goal of that PR was to import the library the same way any user would, so I guess test files should be updated to `import {...} from '@testing-library/vue`.

Not sure how Jest handles scoped packages, though. But happy to look into that if you think that'd help. I guess the index file could be renamed to `index.js` again, and map "@testing-library/vue" to it.

dfcook:
That makes sense, I'm holding back merging this in as currently unable to publish the scoped package on npm (have a ticket out for that).

If you want to take a look at fixing either of the issues you raised in the meantime I'd be happy to accept PR's for those. I'd get to them eventually but work issues have meant I haven't spent as much time working on this as I'd like recently.

Issue #27, opened 2019-05-29
Provide a cleanup-after-each file?


afontcu:
Hi! `react-testing-library` is providing a [cleanup-after-each](https://github.com/testing-library/react-testing-library/blob/master/cleanup-after-each.js) file that comes in handy to avoid repeating `afterEach(cleanup)` on every test file. They even load it using a setup file, so you load it once and forget about dirty DOM nodes once for all.

Would it make sense to port this feature to vue-testing-library? If so, I'd happily submit a PR with both the file and the required documentation 🤗 

👋

dfcook:
Would appreciate thoughts on the PR for this

afontcu:
Done! 🎉

dfcook:
Released in 1.0.1

PR #31, opened 2019-06-01
#27 - Provide cleanup-after-each

dfcook:
1. Add cleanup-after-each in same manner as react-testing-library
2. Add jest module name mapper so tests can import the scoped package
3. Update package name in readme

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 node_modules
 coverage
-npm
+dist
 .opt-in
 .opt-out
 .DS_Store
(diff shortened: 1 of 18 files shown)


dfcook:
Not sure how that file got missed, I've added it now.

I've added cleanup to each test file that had > 1 test. Looking at how the cleanup-after-each is implemented it specifically references dist so not sure it would be runnable from the tests.

afontcu:
Yes, you are right 👍  Just checked `react-testing-library` and they also import and add `afterEach(cleanup)` in their demos ([example](https://github.com/testing-library/react-testing-library/blob/master/examples/__tests__/react-redux.js#L64)). It's OK I think, the `cleanup-after-each` file is a nice addition but the important part is cleaning up the DOM so tests don't mess with each other.

PR #32, opened 2019-06-01
Issue/27 cleanup after each

dfcook:


afontcu:
Oh, damn. Not sure how, but could we write some tests to make sure the file is properly exported (and that is actually cleans up after each test)?

Issue #22, opened 2019-05-13
Is there a better way of handling v-model?


afontcu:
Hi! I know that #16 has been closed and that there's an answer to the original question, but I was wondering if there's any advance in the idea of a `fireEvent.update` method.

I guess it'd make sense to come up with a way of handling input data consistently, without depending on `v-model` being there... am I right? For that to happen, should the issue mentioned [here](https://github.com/testing-library/vue-testing-library/blob/master/tests/__tests__/form.js#L13) be fixed? By the way, is there an open issue in the official library repo? Failed to find any reference.

Thanks, and thank you for your work with this library. Testing is pleasant again.

dfcook:
Finally got around to looking at this, PR raised.

Assuming this goes in, should updateState be removed?

afontcu:
> Assuming this goes in, should updateState be removed?

Makes sense! It is quite pointless now, isn't it? Its only purpose AFAIK was to make v-model work.

dfcook:
Fixed in #34

PR #34, opened 2019-06-01
Add update method for model updates

dfcook:
Add update method to handle form updates without having to call updateState

### DIFF
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -73,7 +73,6 @@ function render (TestComponent, {
       wrapper.setProps(_)
       return wait()
     },
-    updateState: _ => wrapper.setData(_),
     ...getQueriesForElement(wrapper.element.parentNode)
   }
 }
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -103,6 +103,42 @@ fireEvent.touch = async (elem) => {
   await fireEvent.blur(elem)
 }
 
+fireEvent.update = async (elem, value) => {
+  const tagName = elem.tagName
+  const type = elem.type
+
+  switch (tagName) {
+    case 'OPTION':
+      elem.selected = value
+
+      const parentElement = this.element.parentElement.tagName === 'OPTGROUP'
+        ? this.element.parentElement.parentElement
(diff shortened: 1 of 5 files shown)


dfcook:
references #22

afontcu:
Looks good! 

This made me wonder about the use cases for `fireEvent.input` or `fireEvent.change` - now that `fireEvent.update` handles everything (am I right?).

Should then `fireEvent.update` be the default way to go when interacting with inputs? It's great that it handles v-model, but I'm a bit torn that the API differs a bit from the standard events (input, change).

src/vue-testing-library.js
```diff
@@ -103,6 +103,42 @@ fireEvent.touch = async (elem) => {
...
+        return fireEvent.input(elem)
+      }
+
+    case 'TEXTAREA':
+      elem.value = value
+      return fireEvent.input(elem)
+
+    case 'SELECT':
+      elem.value = value
+      await fireEvent.change(elem)
```
afontcu: quick question! Should this line `return`, instead of `await`? If not, why does it need to be different?

dfcook: I missed that, updated now.

WRT change and input, I hadn't considered that update may replace them all. I would rather expose them and allow the user the choice of which to use than deprecate them. What do you think?

afontcu: This actually leads me to another question... 

I feel that just writing `.input` or `.change`, like the native events, makes a lot of sense. And it would be quite weird if `fireEvent.input` only worked in certain cases (when v-model is not used).

So... what if `input` and `change` would handle v-model properly, instead of creating a new event? Would it be hard?

dfcook: We could mask input or change, however we're still moving away from what is the idiomatic usage and taking away options from the end user (to use input or change as they are natively).

I'm not sure if there is an ideal solution here tbh, although I think update as it is here is probably the most flexible solution.

afontcu: You are right. Also adding a method is a non breaking change, while modifying the implementation of `.input` and `.change` might be. Let's add the update method and see how it goes!

PR #38, opened 2019-06-03
Signup

afontcu:
Hi! I noticed % code coverage descended a bit with the introducion of `fireEvent.update`, mostly because we didn't have tests for each element.

I added a few in this PR (to make it look more "natural", I switched from Login form to a Signup one, when it kinda makes more sense to have additional form elements).

However, some use cases are still missing (`<input type="radio">`, for instance). So my question is: should I add all use cases in the same component? Or is it already too large? The Signup form is quite simple (just handles inputs and emits...) but it might become too much at some point. 

Any thoughts on this? 😃


dfcook:
I think it's better to split it out or come up with a better example than a signup form if we're going to add new components to it.

afontcu:
> I think it's better to split it out or come up with a better example than a signup form if we're going to add new components to it.

Fair enough! What about a "Create Review" form?

Title -> input text
Review -> textarea
Rating -> radio button
Genre -> select
Would you recommend? -> checkbox
Submit -> input submit

We might need something for optgroups too 🤔 


Thanks!

afontcu:
Ok, so I updated the PR! 😃 

- Renamed old `Form` component to `Stubs`, to match the test filename.
- Changed to example from a contrived signup form to an Add Review form.


Do you feel it has improved? Any other ideas?

afontcu:
Glad you liked it! I'll merge this, but feel free to suggest any potential improvement! I'll happily update it.

PR #55, opened 2019-07-08
Fix radio, checkbox and select

afontcu:
AFAIK, `checkbox` and `radio` elements use a `checked` property to express that they've been selected, which is set to `true`. So there's no need to pass a `value` argument.

Same thing with `<option>` elements. There was an issue with `this` (due to copy&paste, I guess! 😇), so I also improved naming a bit. 

I also improved tests a bit, to demonstrate that the updated element in each case is properly being checked.


afontcu:
btw, I tried to update to the newest version of `jest-dom/extend-expect`, which includes a `.toHaveValue()`,  but I couldn't make tests pass. I'm not sure if it's on us or on them, but tests look fine right now... 🤔

Issue #49, opened 2019-06-29
Add additionalOptions test example


afontcu:
Adding a test example for `additionalOptions` feature (#48 #47) would both (1) show how to use it and (2) increase code coverage.

I guess using `vue-i18n` as a example would be great.

makeupsomething:
I can work on this as I actually implemented myself today. Not sure if I did it correctly or not though.

```js
import Vue from 'vue'
import Vuei18n from 'vue-i18n'
import ja from '@/translations/ja'

Vue.use(Vuei18n)

const i18n = new Vuei18n({
    locale: 'en',
    messages: {
        ja,
    },
})

import {cleanup, render, fireEvent} from '@testing-library/vue'

afterEach(cleanup)

test('can navigate to signup page from welcome page', async () => {
    const {queryByTestId, getByText, debug} = render(
        App,
        {
            routes,
            store,
            i18n,
        },
        vue => vue.use(Vuei18n),
    )
})
```

If theres a better way let me know and I can tidy this up and add to the tests

afontcu:
Hi! Thank you for taking this :)

Notice how, in your snippet, you are using an external Vue instance to register i18n, so you are not actually using the instance exposed by Vue Testing Library. Notice how, in issue https://github.com/testing-library/vue-testing-library/issues/47, the example suggests returnint `{ i18n }` from the callback function. This is the feature we added on #48.

Long story short: You shouldn't import and use external Vue, but rather use the Vue instance provided in the callback function. Then, create the config there, and return the desired object keys.

If you want, open up a PR and we can work from there :)

Thanks!

makeupsomething:
Hi! 

Yes, I was sure that I should not be importing the external Vue to create the i18n instance but could not figure out how to create it any other way. I had not seen #47 at the time but that is clearly the correct way to create it. 

Il adjust things and create a PR over the weekend.

afontcu:
Closed in #58

PR #58, opened 2019-07-17
Add additionalOptions test example

makeupsomething:
I added an example of using additional options specifically using it with the vue-i18n library

#49


tests/__tests__/vueI18n.js
```diff
@@ -0,0 +1,29 @@
...
+    const i18n = new Vuei18n({
+      locale: 'en',
+      messages: {
+        ja
+      }
+    })
+    return { i18n }
+  })
+
+  expect(queryByTestId('section-header')).toHaveTextContent('Hello')
```
afontcu: I'd suggest using `getByText` instead of `*ByTestId`), since it is the suggested way to go (`data-testid` should be considered a escape hatch for situations where no other query can yield the desired result).

So, for example:

```suggestion
  getByText('Hello')
```

or even

```suggestion
  expect(getByText('Hello')).toBeInTheDocument()
```

(both are the same).

Same thing in other queries.

tests/__tests__/vueI18n.js
```diff
@@ -0,0 +1,29 @@
...
+test('can render en and ja text in header', async () => {
+  const { queryByTestId, getByTestId } = render(VueI18n, {}, vue => {
+    vue.use(Vuei18n)
+    const i18n = new Vuei18n({
+      locale: 'en',
+      messages: {
+        ja
+      }
+    })
+    return { i18n }
```
afontcu: Could we add a comment above this line? Explaining how the returned object is gonna end up merged with the component mounted options (since this is the point of the example).

tests/__tests__/vueI18n.js
```diff
@@ -0,0 +1,29 @@
...
+  })
+
+  expect(queryByTestId('section-header')).toHaveTextContent('Hello')
+
+  await fireEvent.click(getByTestId('button-ja'))
+
+  expect(queryByTestId('section-header')).toHaveTextContent('こんにちは')
```
afontcu: We could also expect that "hello" is not in the document anymore:

```js
expect(queryByText('Hello')).toBeNull()
```

Issue #68, opened 2019-08-03
Add <slot> testing demo


afontcu:
Even though there are no differences between VTL and @vue/test-utils regarding slots, I believe having an example would harm noone.

It'd be great if the example file:
* Used a simple, default slot.
* Used named slots.
* Used scopes slots.
* Stated that there should be no differences between testing slots in VTL and VTU.

afontcu:
Closed in #71

PR #71, opened 2019-08-08
#68 add example for slots and scoped slots

mb200:
For #68. Adds a simple `Card` component and test that asserts slot and scopedSlot content.

### DIFF
--- a/tests/__tests__/slots.js
+++ b/tests/__tests__/slots.js
@@ -7,7 +7,7 @@ import Card from './components/Card'
 // Usage is the same as Vue Test Utils, since slots and scopedSlots
 // in the render options are directly passed through to the Utils mount().
 // For more, see: https://vue-test-utils.vuejs.org/api/options.html#slots
-test('Card component', async () => {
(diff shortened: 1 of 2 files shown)


tests/__tests__/slots.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+  })
+
+  // The scoped prop "content" should be rendered in the given template above.
+  getByText('Yay! Scoped content!')
+
+  // And the header and footer slots should be rendered with the given templates.
+  getByText('HEADER')
+  getByText('FOOTER')
+})
```
afontcu: Could we assert that "Nothing used the Scoped content!" is not in the DOM?

I know this is almost as testing Vue slots system, but given that this is a demo file, I'd like people to fully understand what happened when we rendered the component.

This is a good opportunity to showcase `.not.toBeInTheDocument()` from `extend-expect`, too.

mb200: Sounds good, I'll add that.

PR #74, opened 2019-08-09
Add typings

### LINKED ISSUE #60: Npm installation error (VueJS/Typescript)
Bug: 

On a VueJS project, using the npm install command and basic import statement ( import { render, fireEvent, cleanup } from '@testing-library/vue']) raises an error:

- Could not find a declaration file '@testing-library/vue'

The error recommends installing @types/testing-library__vue. But attempting to install this provides another error message:

-  @types/testing-library__vue@latest is not in the npm registry. 

To reproduce:

- Spin up a vue app with Jest as the default testing provider and use typescript
- Open a .spec.ts file
- Run the npm install for vue testing library 
- Add import { render, fireEvent, cleanup } from '@testing-library/vue';
- IDE picks out import library as an error, npm run test will no longer run successfully
- Installing the suggested library @types/testing-library__vue --save-dev is unsuccessful 

Related information:
- `@testing-library/vue` version: ^1.1.0
- `Vue` version: ^2.6.10
- `node` version: 10.15.3
- `npm` version: 6.9.0

Related code/config:
- Using typescript

cimbul:
Followed recommendations in [dtslint tool](https://github.com/microsoft/dtslint) and [TypeScript documentation](http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html).

Took tsconfig.json and tslint.json from @testing-library/dom, although it looks like they just copied the examples in the dtslint docs.

Closes #60.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1226,6 +1226,12 @@
       "integrity": "sha512-f5j5b/Gf71L+dbqxIpQ4Z2WlmI/mPJ0fOkGGmFgtb6sAu97EPczzbS3/tJKxmcYDj55OX6ssqwDAWOHIYDRDGA==",
       "dev": true
     },
+    "@types/parsimmon": {
+      "version": "1.10.0",
+      "resolved": "https://registry.npmjs.org/@types/parsimmon/-/parsimmon-1.10.0.tgz",
+      "integrity": "sha512-bsTIJFVQv7jnvNiC42ld2pQW2KRI+pAG243L+iATvqzy3X6+NH1obz2itRKDZZ8VVhN3wjwYax/VBGCcXzgTqQ==",
+      "dev": true
+    },
     "@types/stack-utils": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/@types/stack-utils/-/stack-utils-1.0.1.tgz",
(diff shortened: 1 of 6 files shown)


afontcu:
Thank you **very** much for this!

I think we should wait before merging this since there are several substantial changes incoming (such as #75 and the DOM Testing Library version bumping).

Moreover, I just noticed [this comment](https://github.com/testing-library/dom-testing-library/pull/335#issuecomment-519955442) regarding typings and its maintainability. I have little to no experience with TypeScript, so I wouldn't be able to keep typings up to date. What are your thoughts about the proposed solution by Kent?

cimbul:
No problem. Yeah, we could put this on DefinitelyTyped if that's the direction that dom-testing-library is going. It doesn't change the maintainability situation, but it does mean you won't be responsible for fielding issues for it.

For what it's worth, since these definitions only express the interface, changes like #75 don't impact them much. The only relevant change I can see is that the main `fireEvent()` function now returns a promise. These types don't use many advanced TypeScript features, but the ones for dom-testing-library do (BoundFunction being one), so I can see how that might make things harder.

afontcu:
For reference, typings from DTL and RTL are moving to DefinitelyTyped:

https://github.com/testing-library/dom-testing-library/pull/337
https://github.com/testing-library/react-testing-library/pull/437

nickmccurdy:
The TypeScript team recommends that packages not written in TypeScript have their types in DefinitelyTyped (`@types` on npm) instead of in the same package. https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

That being said, you should be able to reuse most of this if you'd like to make a pull request to DefinitelyTyped, though they have some additional tooling and code style requirements.

cimbul:
I'll submit the DefinitelyTyped PR once the dom-testing-library typings get moved there. (Otherwise it's a whole process to add `@testing-library/dom` to a whitelist, only to then remove it.) I'll close this and add a comment when I get that PR submitted.

afontcu:
Thanks @cimbul ❤️

cimbul:
DefinitelyTyped PR here: DefinitelyTyped/DefinitelyTyped#37655

PR #75, opened 2019-08-10
Use and expose a local copy of fireEvent

afontcu:
As suggested in https://github.com/testing-library/user-event/pull/153, VTL mutating `fireEvent` from DOM Testing Library is giving us a hard time trying to make `user-event` usable in VTL. Also, I'd say that mutating an imported object isn't the safest strategy.

I've also added a bunch of tests for the "new" `fireEvent`.

~This is not a breaking change.~ Actually, it is. `fireEvent()` now returns a Promise, and it didn't.

### DIFF
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -92,7 +92,9 @@ function cleanupAtWrapper(wrapper) {
 }
 
 // Vue Testing Library's version of fireEvent will call DOM Testing Library's
-// version of fireEvent plus wait for one tick of the event loop so that...
+// version of fireEvent plus wait for one tick of the event loop to allow Vue
+// to asynchronously handle the event.
+// More info: https://vuejs.org/v2/guide/reactivity.html#Async-Update-Queue
 async function fireEvent(...args) {
   dtlFireEvent(...args)
   await wait()
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -4,7 +4,7 @@ import {
   getQueriesForElement,
   prettyDOM,
   wait,
-  fireEvent
+  fireEvent as dtlFireEvent
 } from '@testing-library/dom'
 
 const mountedWrappers = new Set()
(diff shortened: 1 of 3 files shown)


src/vue-testing-library.js
```diff
@@ -91,10 +91,16 @@ function cleanupAtWrapper(wrapper) {
   mountedWrappers.delete(wrapper)
 }
 
-Object.keys(fireEvent).forEach(fn => {
-  fireEvent[`_${fn}`] = fireEvent[fn]
-  fireEvent[fn] = async (...params) => {
-    fireEvent[`_${fn}`](...params)
+// Vue Testing Library's version of fireEvent will call DOM Testing Library's
+// version of fireEvent plus wait for one tick of the event loop so that...
```
afontcu: "so that..."

couldn't think of a good wording to explain why waiting for a tick is needed. Any ideas?

afontcu: I'd love to know, too, whether if making fireEvent return a promise is totally necessary due to Vue internals. There might be some alternative that keeps VTL public API as similar as possible to RTL's and DTL's.

calebeby: >Any ideas?

How about "to allow vue to asynchronously handle the event"?

Maybe it would make sense to link to the docs here: https://vuejs.org/v2/guide/reactivity.html#Async-Update-Queue

Good question on if it is necessary for fireevent to be async. Both React and Preact expose ways of forcing them to flush their update queues synchronously. Maybe it is worth looking into [`vm.$forceUpdate`](https://vuejs.org/v2/api/#vm-forceUpdate) to see if it does what we want

afontcu: > How about "to allow vue to asynchronously handle the event"?

Sounds good! Added.

I'm gonna give `vm.$forceUpdate` (or even `nextTick`? who knows) a go. I'd really love to stick to synchronous events.

calebeby: Might also be worth looking into https://github.com/vuejs/vue-test-utils/issues/1137

afontcu:
Thanks for the approval @dfcook! Merging this one. Are you publishing a new major version?

dfcook:
2.0.0 released

PR #80, opened 2019-08-10
automatically clean up aftereach with optout using VTL_SKIP_AUTO_CLEANUP

### LINKED ISSUE #77: Automatically cleanup if afterEach is detected
Right now, users need to call `afterEach(cleanup)` or configure Jest [to do so](https://testing-library.com/docs/vue-testing-library/setup#cleanup). This makes sure test run in isolation, but requires manual action.

---

As React Testing Library [just did](https://github.com/testing-library/react-testing-library/pull/430), we should probably automatically call `cleanup` if `afterEach` method exists.

RTL ended up offering two opt-out mechanisms (via custom import or using an env variable), but I'm not sure this is totally needed. I would keep the env variable (`VTL_SKIP_AUTO_CLEANUP`), but still have a single import point (instead of creating that "pure" version on VTL). Not a big deal, though.

So a PR (or several PRs) should:
* Automatically call `afterEach(cleanup)` if `afterEach` exists.
* Add an opt-out mechanism by using `VTL_SKIP_AUTO_CLEANUP`.
* Add tests asserting the behavior outlined above.
* Remove `cleanup` references from tests and docs.
* Replace cleanup-after-each with a warning ([reference](https://github.com/testing-library/react-testing-library/commit/8277b06c199cacb0e81597aa506151de7a71b439)).
* State that the change is a breaking change (so that we take it into account when merging!).

After that, we'd need to update [the official docs](https://github.com/testing-library/testing-library-docs/).

I'm marking this as a good first issue since details are outlined and the referenced PR from RTL makes a great starting point :)

Oluwasetemi:
Closes #77

- [x] Automatically call afterEach(cleanup) if afterEach exists.
- [x] Add an opt-out mechanism by using VTL_SKIP_AUTO_CLEANUP.
- [x] Add tests asserting the behavior outlined above.
- [x] Remove cleanup references from tests and docs.
- [x] Replace cleanup-after-each with a warning.
- [x] State that the change is a breaking change.


dfcook:
This looks good, can you add a couple of tests illustrating the usage of VTL_SKIP_AUTO_CLEANUP?

Oluwasetemi:
I will add the test to it right away

tests/__tests__/auto-cleanup-skip.js
```diff
@@ -0,0 +1,20 @@
+import '@testing-library/jest-dom/extend-expect'
```
afontcu:
```suggestion
```

`extend-expect` is not in use in the test, so we can skip importing it

tests/__tests__/auto-cleanup.js
```diff
@@ -0,0 +1,26 @@
...
+
+  const h1 = container.querySelector('h1')
+  expect(h1.textContent).toEqual('Hello World')
+})
+
+const NewComponent = {
+  template: `<h1>Automatic clean after each test</h1>`
+}
+
+test('it should cleanup after each test and render NewComponent', () => {
```
afontcu:
```suggestion
test('cleans up after each test by default', () => {
```

tests/__tests__/auto-cleanup.js
```diff
@@ -0,0 +1,26 @@
+import { render } from '@testing-library/vue'
+import '@testing-library/jest-dom/extend-expect'
```
afontcu: same here (even though here we could use `.toHaveTextContent()` instead of manually checking `textContent`)

tests/__tests__/auto-cleanup.js
```diff
@@ -0,0 +1,26 @@
...
+  expect(h1.textContent).toEqual('Hello World')
+})
+
+const NewComponent = {
+  template: `<h1>Automatic clean after each test</h1>`
+}
+
+test('it should cleanup after each test and render NewComponent', () => {
+  render(NewComponent)
+  expect(document.body.innerHTML).toEqual(
```
afontcu: I believe we could keep this a bit simpler. Instead of rendering a new component, we could simply expect that the `<body>` is empty:

```js
expect(document.body.innerHTML).toEqual('')
```

tests/__tests__/auto-cleanup-skip.js
Oluwasetemi: ok. I will remove it.

Oluwasetemi:
Can you please take a look at this!

tests/__tests__/auto-cleanup.js
```diff
@@ -0,0 +1,19 @@
...
+import '@testing-library/jest-dom/extend-expect'
+
+const HelloWorldComponent = {
+  template: `<h1>Hello World</h1>`
+}
+// This just verifies that by importing VTL in an
+// environment which supports afterEach (like jest)
+// we'll get automatic cleanup between tests.
+test('render the first component', () => {
+  const { container } = render(HelloWorldComponent)
```
afontcu: I like how you inlined the component in the `auto-cleanup-skip.js` test. Could we do the same thing here, instead of creating a "global" `HelloWorldComponent`?

Oluwasetemi:
Done!!!

tests/__tests__/auto-cleanup.js
Oluwasetemi: Done!

afontcu:
Hi @Oluwasetemi! This looks awesome, thank you 🙌 

I was making sure everything was in place, and came up with an idea. 

What if we used `toMatchInlineSnapshot()` ([docs](https://jestjs.io/docs/en/snapshot-testing#inline-snapshots)) to generate the snapshots to check against?

```js
// auto-cleanup-skip.js

test('first test render a vue component', () => {
  render({
    template: `<h1>Hello World</h1>`
  })

  expect(document.body.innerHTML).toMatchInlineSnapshot(`
    <div>
      <h1>Hello World</h1>
    </div>
  `)
})

test('no cleanup should have happened, renders the first component still', () => {
  expect(document.body.innerHTML).toMatchInlineSnapshot(`
        <div>
          <h1>Hello World</h1>
        </div>
    `)
})
```


```js
// auto-cleanup.js

test('render the first component', () => {
  render({
    template: `<h1>Hello World</h1>`
  })

  expect(document.body.innerHTML).toMatchInlineSnapshot(`
                <div>
                  <h1>Hello World</h1>
                </div>
        `)
})

test('cleans up after each test by default', () => {
  expect(document.body.innerHTML).toMatchInlineSnapshot(`""`)
})
```

What do you think about it? I just wrote `expect(document.body.innerHTML).toMatchInlineSnapshot()` and let Jest create the initial version. Looks like a good use case for snapshot testing!

Oluwasetemi:
It definitely looks better. This is way better than the previous implementation of the tests.

It will be visible to anyone who checks through the test to understand the cleanup and auto clean up properly.

Oluwasetemi:
How will the docs be updated?

afontcu:
> How will the docs be updated?

Docs live in https://github.com/testing-library/testing-library-docs/, so a PR should be created there with the appropriate changes.

afontcu:
Just a small note: we are exploring ways of automating CI/CD (see #85 #86), and I believe this PR is a great way of checking the new workflow. This is why is not merged yet, even though it looks great and all work is done 👍

PR #86, opened 2019-08-17
Move to kcd-scripts

afontcu:
By using `kcd-scripts` we stay one step closer to Testing Library sibling packages while removing the need to handle stuff manually.

- [x] Installed `kcd-scripts` to manage npm scripts.
- [x] Customized them to work with Vue files.
- [x] Move test files to `src/__tests__` to honor default config.
- [x] Achieved 100% code coverage (required by default kcd-scripts config). I had to remove a seemingly useless if statement [here](https://github.com/testing-library/vue-testing-library/compare/kcd-scripts?expand=1#diff-2cfdd145f61e8456ea199f8af800c145R57). Correct me if I'm wrong!
- [x] Ran `npm run format` (this is why there's a lot of small styling changes - mostly trailing commas and whitespaces around brackets).
- [ ] Test travis + coveralls + semantic-release integration. Not sure how to do that.

**Is this a Breaking Change?**:
I'd love to make sure that the `build` output is still okay. I removed `browsers` from `.babelrc` to use the default config (the same used in RTL, for instance).


afontcu:
I''ll shamelessly ping @kentcdodds to see if he can provide some insights with regarding testing the Travis part.

🙌

package.json
```diff
@@ -80,11 +68,12 @@
   },
   "lint-staged": {
     "*.{js,vue}": [
-      "eslint --fix",
+      "kcd-scripts format",
```
kentcdodds: You can get rid of the lint-staged config entirely and add husky config instead like my other projects and that do the same thing

kentcdodds:
As far as testing, it's hard to test Travis. I normally just do trial and error 😅

package.json
afontcu: I tried to, but kcd-scripts' `.lintstagedrc.js` ([source](https://github.com/kentcdodds/kcd-scripts/blob/master/src/config/lintstagedrc.js#L8)) wouldn't get trigged by `.vue` files 😢 

Would you accept a PR to add them? That wouldn't make kcd-scripts to work with Vue files out-of-the-box, but wouldn't hurt anybody either.

ps: thanks for the feedback! 💯

kentcdodds: Sure 👍

package.json
```diff
@@ -1,22 +1,28 @@
 {
   "name": "@testing-library/vue",
-  "version": "2.0.0",
+  "version": "0.0.0-semantically-released",
   "description": "Simple and complete Vue DOM testing utilities that encourage good testing practices.",
   "main": "dist/vue-testing-library.js",
```
kentcdodds: I believe this will need to be changed

afontcu: Oh, how come? I was expecting Travis + semantic release will set the version right before it releases.

Actually, I wanted to ask you something. I saw that `semantic-release` uses git commit messages to compute the required version bump. I've noticed, however, that you are not enforcing any commit structure in RTL, for instance ([example](https://github.com/testing-library/react-testing-library/pull/427)). How does that work?

afontcu: Oh, sorry Kent, I thought you mean the `version` field, now I see you added your comment two lines below. I double checked, and since the file is called `src/vue-testing-library.js`, the generated dist/ folder keeps the same naming.

kentcdodds: When I merge pill requests I use the "Squash and merge" option which allows me to write my own commit message. Much less of a burden on contributors and I don't need commitizen.

.babelrc
```diff
@@ -2,18 +2,9 @@
   "presets": [
     ["@babel/env", {
       "modules": "commonjs",
```
afontcu: @dfcook do you believe this ` "modules": "commonjs",` statement necessary, given the environment VTL runs on?

dfcook: No, probably not, feel free to change

afontcu:
> This is great, do we have a workflow for commit -> release now?

So the idea is that we set VTL repo to only work with "squash and merge" strategy (this can be done through config, no problem), so that **we get to author the commit message that merges to master**. This way, by following `semantic-release` [guidelines](https://semantic-release.gitbook.io/semantic-release/#commit-message-format), Travis script will automatically run tests, update to the appropriate package version, and deploy it.

At least this is the idea, can't (obviously) test it until everything gets merged and we see the results 😇 

If that sounds reasonable, I'd say we merge this as a patch version - in this case, I'd use a commit message that reads "chore: move building process to kcd-scripts" - and see how it goes. A lot of things (github/npm tokens, config, etc) could go wrong, but we have to give it a go :)

afontcu:
Okay I'm giving this a go. Wish me luck 😂

afontcu:
Okay, once I created the appropriate git tag (`v2.0.1`) to match the current published version, looks like the script is failing at `npm publish` (after properly inferring the next version and so on).

```c
[release] npm notice 
[release] npm ERR! code E404
[release] npm ERR! 404 Not Found - PUT https://registry.npmjs.org/@testing-library%2fvue - Not found
[release] npm ERR! 404 
[release] npm ERR! 404  '@testing-library/vue@2.1.0' is not in the npm registry.
```

It is weird because the [URL actually exists](https://registry.npmjs.org/@testing-library%2fvue) and the library is obviously published 🤷‍♀ 

imatge

kentcdodds:
You have the GH_TOKEN and NPM_TOKEN set right?

afontcu:
NPM_TOKEN was created/uploaded when I ran `semantic-release-cli setup`, but I just noticed that I'm not "owner" of VTL in npm (just checked by running `npm owner ls @testing-library/vue`). That should make a difference, right?

kentcdodds:
I just created and added an `NPM_TOKEN` using the `testing-library-bot` account which should have access. We should be good to go now.

afontcu:
That was it @kentcdodds 💯💯💯💯 thank you!!

weyert:
Cool! Great job guys! Does anyone know if you can zennify the squash and merge commit message, though?

kentcdodds:
I don't know if any way to automate it, but I don't see any reason to either 🤷‍♂️

weyert:
I was just wondering, we are switching to Github at work and if you squash and merge allows to change the commit message. Guess it would be nice to avoid fix it messages 😀

kentcdodds:
Yep, it does allow you to change the message when you squash and merge 👍

PR #100, opened 2019-09-30
Bound queries to the base element

### LINKED ISSUE #98: Provide options to adapt the document before mount
**Describe the feature**
We are using Vuetify as component library in one of our projects. I would like to introduce vue-testing-library and came up with some setup troubles on the way. Vuetify requires a wrapper element containing the attribute `data-app`. This element is used to append components like dialog to the root DOM level. The required setup can be done using jest setup files. Thought, the dialog content can not be found when using the exposed queries of the vue-testing-libraries `render()` method.

**To Reproduce**

- Init a vue app with vuetify
- Create a basic dialog
- write a test which should validate that the dialog content is visible when a button is clicked
- Add Vuetify and `<div data-app></div>` to your test DOM instance before mount

=> queries like getByText will not query the dialog content

I tried to create a codesandbox, but failed. Is there a working example for vue-testing-library?
https://codesandbox.io/s/vuetify-with-vue-testing-library-bhu8f?previewwindow=tests

**Expected behavior**
I expect a configuration option like it exists for react:

- baseElement: Specify the base element for the queries.
- container: Specify the surrounding DOM node

**Related information:**

- `@testing-library/vue` version: 3.0.0
- `Vue` version: 2.5.18
- `node` version: 8.12.0
- `yarn` version: 1.17.3
- `Vuetify` version: 1.2.10

```javascript
import CreateLorem from "./CreateLorem";
import { render } from "@testing-library/vue";
import Vuetify from "vuetify";
import Vue from "vue";

var app = document.createElement("div");
app.setAttribute("data-app", true);
document.body.appendChild(app);
Vue.use(Vuetify);

it("should create a lorem", () => {
  const { getByText } = render(CreateLorem);
  getByText("open");
});
```

afontcu:
This PR aims to fix several issues when trying to query content outside of the regular DOM node tree (portals, modals, etc). It tries to mirror React Testing Library behavior.

I also added a test based on Vuetify, because (1) it is related to the introduced changes and (2) Vuetify is widely-used and comes with its own nuances.

This is a **breaking change** since `container` and `baseElement` changed their meaning.

This PR closes #98

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -1,0 +1,36 @@
+import {render} from '@testing-library/vue'
+import '@testing-library/jest-dom/extend-expect'
+
+test('returns baseElement which defaults to document.body', () => {
+  const {baseElement} = render({template: '<div />'})
+  expect(baseElement).toBe(document.body)
+})
+
+test('returns custom baseElement', () => {
+  const {baseElement} = render(
+    {
+      template: '<div />',
+    },
+    {
+      baseElement: document.createElement('blink'),
+    },
+  )
+
+  expect(baseElement).toMatchInlineSnapshot(`
+    <blink>
+      <div>
+        <div />
+      </div>
+    </blink>
+  `)
+})
+
+test('renders container', () => {
+  const {container, getByTestId} = render({
+    template: '<div data-testid="myDiv">my content</div>',
+  })
+
+  expect(container.firstChild).toHaveTextContent(
+    getByTestId('myDiv').textContent,
+  )
+})
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -1,27 +1,30 @@
 import {render} from '@testing-library/vue'
 import '@testing-library/jest-dom/extend-expect'
 
-test('returns baseElement which defaults to document.body', () => {
+test('baseElement defaults to document.body', () => {
   const {baseElement} = render({template: '<div />'})
   expect(baseElement).toBe(document.body)
 })
 
-test('returns custom baseElement', () => {
-  const {baseElement} = render(
-    {
-      template: '<div />',
-    },
-    {
-      baseElement: document.createElement('blink'),
-    },
-  )
+test('renders custom baseElement', () => {
+  const Component = {template: '<span />'}
+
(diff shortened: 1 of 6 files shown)


src/vue-testing-library.js
```diff
@@ -53,15 +57,12 @@ function render(
   })
 
   mountedWrappers.add(wrapper)
-
-  const div = document.createElement('div')
-  wrapper.element.parentNode.insertBefore(div, wrapper.element)
```
afontcu: I'm not 100% sure this line should be gone. If I add it back, [this inline snapshot](https://github.com/testing-library/vue-testing-library/pull/100/files#diff-b45a2abc5c6c583db5549af5a49a2aa7R19) changes to `<blink></blink>`, and I don't think that's the desired output.

@dfcook could you double check that the logic here is no longer necessary as long as we append `wrapper.element` to our container (which is, in turn, a `<div>` wrapped in a `<body>`)?

dfcook:
All looks good to me, I think the resulting code is a lot cleaner as well.

src/vue-testing-library.js
dfcook: I agree, given that we are now in control of the container we can remove this logic.

src/__tests__/render.js
```diff
@@ -0,0 +1,36 @@
...
+  expect(baseElement).toBe(document.body)
+})
+
+test('returns custom baseElement', () => {
+  const {baseElement} = render(
+    {
+      template: '<div />',
+    },
+    {
+      baseElement: document.createElement('blink'),
```
ChristophWalter: I really like the new interface. But it is different from the react one which might be misleading. There a `container` is used to create an element and the `baseElement` just acts as pointer to that: https://testing-library.com/docs/react-testing-library/api#container

afontcu: Hm, I see. I kind of like our API better, but it's true that keeping things as simillar as possible is a must. Thanks for pointing it out!

afontcu:
@ChristophWalter @dfcook  I've updated the PR in order to make it similar to React Testing Library.

https://testing-library.com/docs/react-testing-library/api#container

PR #110, opened 2019-11-04
add simple directive and test

### LINKED ISSUE #108: add example for using directives
as per https://spectrum.chat/testing-library/help-vue/mocking-directives~1167b8e5-a303-494e-a0c5-593d75e8b792 example is needed to test a component that is using a directive

chiptus:
fixes #108

### DIFF
--- a/src/__tests__/components/CompWithDirective.vue
+++ b/src/__tests__/components/CompWithDirective.vue
@@ -12,5 +12,3 @@ export default {
   },
 }
 </script>
-
-<style lang="scss" scoped></style>
--- a/src/__tests__/components/CompWithDirective.vue
+++ b/src/__tests__/components/CompWithDirective.vue
@@ -1,0 +1,16 @@
+<template>
+  <input v-value="value" />
+</template>
+
+<script>
+export default {
+  props: {
+    value: {
+      type: String,
+      required: true,
+    },
+  },
+}
+</script>
+
+<style lang="scss" scoped></style>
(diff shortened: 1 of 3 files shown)


src/__tests__/components/CompWithDirective.vue
```diff
@@ -0,0 +1,16 @@
...
+  props: {
+    value: {
+      type: String,
+      required: true,
+    },
+  },
+}
+</script>
+
+<style lang="scss" scoped></style>
```
afontcu:
```suggestion
```

src/__tests__/use-global-directive.js
```diff
@@ -0,0 +1,18 @@
+import '@testing-library/jest-dom/extend-expect'
+import {render, wait} from '@testing-library/vue'
+import {value as valueDirective} from './components/directives/v-value'
+import CompWithDirective from './components/CompWithDirective.vue'
+
+test('element should have the same value as the directive', async () => {
```
afontcu: I feel the example is a bit "wordy" (it says `value` a lot 😝).

I think it would be best if we used a simpler example with a clearer directive. Also I'd love to avoid using `updateProps`, as it has actually nothing to do with directives, and people may get the wrong impression.

What do you think about that?

chiptus:
this was closed because I referenced it in #111, I'll fix the issues and reopen

afontcu:
Hi @chiptus! Any chance you could take a look at this? 😄 

have a happy new year!

chiptus:
will do

afontcu:
Fixed in #120. Thanks for this, though! 😃

Issue #122, opened 2020-03-04
store not available in child components


mediafreakch:
**Describe the bug**
`store` is undefined in child components.

**To Reproduce**

1. Clone https://github.com/mediafreakch/vue-testing-library-bug
2. run `npm install`
3. run `npm test`

The second test fails.

**Expected behavior**

Second test passes.

**Screenshots**

**Related information:**

- `@testing-library/vue` version: 4.1.0
- `Vue` version: 2.6.11
- `node` version: 12.13.1
- `npm` (or `yarn`) version: npm 6.12.1

Relevant code or config (if any)

```javascript
```

**Additional context**

Unfortunately the example doesn't run in Codesandbox. There is an issue with they're bundler.
You need to clone the repo and run it locally to reproduce the bug.

I managed to get the same app tested using pure `@vue/test-utils`. ~I will link the corresponding codesandbox example as soon as I have recovered it~. [Codesandbox example](https://codesandbox.io/s/vibrant-aryabhata-sxkfj) of `@vue/test-utils`

mediafreakch:
After all I've read, could this be due to a clash of Babel versions? 

`vue-testing-library: @babel/core@7.8.4`
`vue-test-utils: @babel/core@6.26.0`

Version `7.8.4` includes `dynamic-import-node` plugin, while the latter does not. 
I wouldn't know how to fix though nor if that really can be a factor...

BeeeQueue:
Can confirm this happening, can't write any tests for components that use vuex. 😞 

It's not the babel version. I tried a resolution for it to `7.10.2` and it didn't change anything.

The store is available deep in the private apis `root._vnode.componentInstance.$store`

afontcu:
Hi! 👋 

Just gave the OP repro a go. Not sure why, but removing the first `mockReturnValueOnce` and only running the `"should render the chart component"` test makes it pass. Same thing if I set the `viewInfo` mock per test, instead of relying on `beforeEach`.

In other words – this works:

```js
  beforeEach(() => {
    actions = {
      fetchViewInfo: jest.fn().mockResolvedValue(null),
    }
  })

  it("should render the chart component", async () => {
    store = {
      actions,
      getters: {
        viewInfo: () =>
          jest.fn().mockReturnValueOnce({
            title: "Test Chart",
          }),
      },
    }

    const { findByText } = render(GenericView, {
      store,
      propsData: {
        kind: "Dashboard",
      },
    })

[...]

PR #123, opened 2020-03-06
Alternative approach for integration with Vuetify

mediafreakch:
I suggest an alternative approach to the problem to preserve things like `props, propsData, updateProps` that are applied to the root node only.

### DIFF
--- a/src/__tests__/vuetify.js
+++ b/src/__tests__/vuetify.js
@@ -15,18 +15,13 @@ Vue.use(Vuetify)
 // a <div data-app="true"> node.
 const renderWithVuetify = (component, options, callback) => {
   return render(
-    // anonymous component
-    {
-      // Vue's render function
-      render(createElement) {
-        // wrap the component with a <div data-app="true"> node and render the test component
-        return createElement('div', {attrs: {'data-app': true}}, [
-          createElement(component),
-        ])
-      },
-    },
+    component,
     // for Vuetify components that use the $vuetify instance property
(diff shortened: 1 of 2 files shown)


src/__tests__/vuetify.js
```diff
@@ -44,3 +39,13 @@ test('renders a Vuetify-powered component', async () => {
     </div>
   `)
 })
+
+test('allows changing props', async () => {
+  const {queryByText, updateProps} = renderWithVuetify(VuetifyDemoComponent)
+
+  expect(queryByText('This is a hint')).toBe(null)
```
afontcu: To make it similar to other examples, I'd suggest using jest-dom expectations

```suggestion
  expect(queryByText('This is a hint')).not.toBeInTheDocument()
```

you'll need to add the import statement, too:

```js
import '@testing-library/jest-dom'
```

src/__tests__/vuetify.js
```diff
@@ -44,3 +39,13 @@ test('renders a Vuetify-powered component', async () => {
...
+
+  await updateProps({showHint: true})
+
+  expect(queryByText('This is a hint')).not.toBe(null)
```
afontcu:
```suggestion
  expect(queryByText('This is a hint')).toBeInTheDocument()
```

afontcu:
Thanks for this! :)

mediafreakch:
@afontcu I was wondering if #122 might be somehow related, even though I'm not extending the render method there...

PR #125, opened 2020-03-12
Bump DTL major version 

afontcu:


src/__tests__/stopwatch.js
```diff
@@ -17,7 +17,7 @@ test('updates component state', async () => {
   getByText('Stop')
 
   // Wait for one tick of the event loop.
-  await wait()
+  await wait(() => {})
```
kentcdodds: Note that this is probably not how I would test a stopwatch. I would prefer `jest.useFakeTimers` (in which case you'd need to use `act`).

Maybe this: https://github.com/kentcdodds/react-testing-library-course/blob/5b1a2d10fa140eae1ea36ec47cc3f52d3edaa90d/src/__tests__/unmounting.js or this: https://github.com/kentcdodds/testing-docs-rtl-examples/blob/cf6b0f3c74287b7b6bca745bc28827c8a2c0c4ee/src/card.test.js can help :)

Thanks!

Issue #119, opened 2020-01-14
Vue router configuration not working with mocha-chai


FCalabria:
**Describe the bug**

When trying to use the router option in a mocha + chai project, it fails with `TypeError: VueRouter is not a constructor`

**To Reproduce**

Run unit tests:

https://github.com/FCalabria/testing-library-mocha

The same project with jest, works perfectly.

**Related information:**

- `@testing-library/vue` version: 4.1.0
- `Vue` version: 2.6.10

Changing the way to import the library fixes the bug, but breaks it for jest

```javascript
const VueRouter = require('vue-router') // current code, works with jest
const VueRouter = require('vue-router').default // fix for mocha, breaks jest
```

PR #127, opened 2020-03-13
fix(#119): allow vue-router creation through routes param from both jest and mocha

agualis:
This code fixes [issue](https://github.com/testing-library/vue-testing-library/issues/119).

### DIFF
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -36,7 +36,7 @@ function render(
   }
 
   if (routes) {
-    const VueRouter = require('vue-router')
+    const VueRouter = require('vue-router') || require('vue-router').default
     localVue.use(VueRouter)
     router = new VueRouter({
       routes,
--- a/src/__tests__/vue-router-mocha.js
+++ b/src/__tests__/vue-router-mocha.js
@@ -1,0 +1,16 @@
+import '@testing-library/jest-dom'
+import {render} from '@testing-library/vue'
+
+import About from './components/Router/About.vue'
+
+const routes = []
+test('uses require("vue-router").default when require("vue-router") is undefined (useful for mocha users)', () => {
+  // Test for fix https://github.com/testing-library/vue-testing-library/issues/119
+  jest.mock('vue-router', () => {
+    return undefined
+  })
+
+  expect(() => render(About, {routes})).toThrowError(
+    new TypeError("Cannot read property 'default' of undefined"),
+  )
+})


src/vue-testing-library.js
```diff
@@ -36,7 +36,7 @@ function render(
   }
 
   if (routes) {
-    const VueRouter = require('vue-router')
+    const VueRouter = require('vue-router') || require('vue-router').default
```
agualis: The code is ugly cause the problem is that  in mocha unit tests we need to bundle the code with webpack (mocha-webpack) and that does not play well with using a require here. 

The problem could be also solved with an async import (to follow ES6 notation) but that would probably make the code even uglier cause it would introduce async code.

Another workaround when using mocha, would be passing an already built router through `router` mounting option (instead  of `routes`)  but that cannot work with the current implementation.

Enabling this option for mocha users would be highly appreciated. Thanks for your awesome work!

afontcu:
Hi! Thanks for this! 😃 

* Is there a way to add some tests to make sure things are working as expected? I'm not well versed with mocha, but that would be great (right now CI pipeline is failing due to code coverage descending).

* Should we do the same thing with Vuex?

agualis:
> Is there a way to add some tests to make sure things are working as expected? I'm not well versed with mocha, but that would be great (right now CI pipeline is failing due to code coverage descending).

Some mocha tests could be added but that would not affect the code coverage as mocha does not use jest runner. 

However, I just commited a _hacky_ jest test  (`vue-router-mocha.js`) that exercises the added code to keep the same coverage ( not the most expressive test though I know 😅)

> Should we do the same thing with Vuex?

No, we don't need it because `vuex` is built/exported in a different way than `vue-router` so that require works out of the box with mocha.

afontcu:
Looks good! I'm willing to merge this 🎉. Yet, do you have an external repository or something where we could see the fix actually fixing the issue? I've never used mocha, so I just want to make sure mocha is 100% usable after this 😄

agualis:
> Yet, do you have an external repository or something where we could see the fix actually fixing the issue? 

The repository pointing the issue (#119) reproduces the error with [this mocha test](https://github.com/FCalabria/testing-library-mocha/blob/master/tests/unit/example.spec.js)

@afontcu If you replace vue-testing-library with the code in current PR, that test will  ✅ 😊

agualis:
BTW, I prefer jest over mocha but it need to use it in a legacy project. I think that I could help to mimic vue-testing-library examples in mocha but I would suggest to do it in the official vue-testing-library repository so it's easier to find/maintain them.

PR #142, opened 2020-05-20
Fix asserting errors in async lifecycle hooks

dlindenkreuz:
Right now, when an async Vue lifecycle hook such as `mounted` throws/rejects, it seems that `expect(cleanup).toThrow()` is the only way to reliably assert an error. When auto-cleanup is enabled, this would mean that `cleanup` is called multiple times for some tests.

`wrapper.destroy`, called by `cleanup`, rethrows any errors (see [@vue/test-utils `throwIfInstancesThrew`](https://github.com/vuejs/vue-test-utils/blob/c94c58985e9e4b82b2c0cde0a1605fb7eaf09fbe/packages/test-utils/src/wrapper.js#L158-L159)). Without try/finally in place, the destroyed wrapper never gets deleted from `mountedWrappers` when `wrapper.destroy` throws.

The assert works, but the test still fails because auto-cleanup throws in `afterEach` as it attempts to destroy the wrapper again. It would be tedious to turn off auto-cleanup for the entire test setup, just to call cleanup manually for all remaining tests.

Note: I added a test that currently creates some `console.error` logs, but none of which affect the test outcome negatively. I currently use jest to stub out `console.error` temporarily. Feel free to remove this if you're fine with some errors being logged.


src/__tests__/cleanup-throw.js
```diff
@@ -0,0 +1,19 @@
...
+    },
+    template: `<h1>Hello World</h1>`,
+  })
+  // thrown errors are logged redundantly by vue-test-utils injected Vue.config.errorHandler
+  // mute console
+  const spy = jest.spyOn(console, 'error').mockImplementation(() => {})
+  await Vue.nextTick()
+  expect(cleanup).toThrow(err)
+  // unmute console
+  spy.mockReset()
```
afontcu, line 18:
```suggestion
  // thrown errors are logged redundantly by vue-test-utils injected Vue.config.errorHandler
  // mute console
  const spy = jest.spyOn(console, 'error').mockImplementation(() => {})

  await Vue.nextTick()
  expect(cleanup).toThrow(err)

  // unmute console
  spy.mockReset()
```

PR #143, opened 2020-05-22
Add a Vue Apollo example

### LINKED ISSUE #141: vue-apollo example
Is there any vue-apollo test example? It would be great to have some guidance on it

cilice:
Closes #141 

I had some freedom the pick some kind of a mocking client for apollo since `vue-apollo` doesn't provide one. 
I also used `screen`, is that alright or should i redo it using the destructing approach used everywhere else?


src/__tests__/vue-apollo.js
```diff
@@ -0,0 +1,55 @@
...
+
+  await waitFor(() => screen.getByText('alice@example.com'))
+  expect(screen.getByText('alice@example.com')).toBeInTheDocument()
+
+  await fireEvent.update(
+    screen.getByLabelText('Email'),
+    'alice+new@example.com',
+  )
+
+  mockClient.setRequestHandler(updateUserMutation, variables =>
```
afontcu, line 43: quick question: Is it needed to define this request handler right here? I was reading the text and felt a bit off to have some "setting up" stuff between the `fireEvent.update` and the `fireEvent.click`

cilice: Oh no, totally not, you can define it right in the beginning, next to the query.
This is also highly specific to the mocking client, usually when you come from React's MockProvider you need to define all queries and mutations beforehand.

cilice:
I guess I should also add a note this is only for Apollo v2 and will not work with Apollo v3.

src/__tests__/vue-apollo.js
afontcu: cool 💯! Then I'd suggest moving it to the beginning of the test, so one can easily spot the "Arrange" section of it :)

src/__tests__/vue-apollo.js
```diff
@@ -0,0 +1,62 @@
...
+
+  await fireEvent.update(
+    screen.getByLabelText('Email'),
+    'alice+new@example.com',
+  )
+
+  await fireEvent.click(screen.getByRole('button', {name: 'Change email'}))
+
+  await waitFor(() => screen.getByText('alice+new@example.com'))
+  expect(screen.getByText('alice+new@example.com')).toBeInTheDocument()
```
afontcu, line 61: question! Can this be simplified to

```suggestion
  await screen.findByText('alice+new@example.com')
```

would it work? Looks like both lines are asserting the same thing (that the text exists)

cilice:
Things I've changed now:
- clarified the setup 
- added some remarks about `apollo-client` v2
- added a loading example and added some comments around the needed `await` for data to appear, since fetching is async

I think this PR is in a good state now, feel free to merge. 🙌 ✨

afontcu:
@cilice I think this comment went out the minute you pushed, you might have missed it xD

https://github.com/testing-library/vue-testing-library/pull/143#discussion_r429185365

src/__tests__/vue-apollo.js
cilice: It does! I've changed it to `expect(await screen.findByText('Email: alice@example.com')).toBeInTheDocument()`

cilice:
Oh thats neat, it somehow never crossed my mind to solve it like this. It's great! I've changed it now and squashed it all into one commit.

cilice:
@afontcu I think it’s ready to merge 🙏

PR #147, opened 2020-06-14
feat: Update to DTL with user-event

afontcu:
_(Not ready to merge)_.

I feel that using `userEvent` means that `fireEvent.update` is no longer needed. However, I  think we should publish this versions as minor one, so that people can start getting used to `userEvent`, and then remove `fireEvent.update` in a future major version.

Missing bits:
- [ ] Wait until https://github.com/testing-library/dom-testing-library/pull/616 is approved, merged, and released.
- [ ] Fix [issues](https://github.com/testing-library/dom-testing-library/pull/616#issuecomment-643814190) with `textarea`.
- [ ] Improve [stopwatch](src/__tests__/stopwatch.js) test. It was already flaky, and this change broke it again.
- [ ] to consider: given that `fireEvent.update` might not be needed anymore, should we introduce a warning to tell people it is deprecated and use `userEvent` instead? (This should come with an opt-out mechanism, too).

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@
 <!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
 
 - [Installation](#installation)
-- [A simple example](#a-simple-example)
+- [A basic example](#a-basic-example)
   - [More examples](#more-examples)
 - [Docs](#docs)
 - [Typings](#typings)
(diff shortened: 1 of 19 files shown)


JacobParis:
Is there anything here you'd like an extra set of hands to tackle? I'm onboarding my team at work to VTL and the sooner user-event is merged the fewer patterns I'll have to reteach

afontcu:
Hi @JacobParis! We unfortunately [hit a ship stopping issue](https://github.com/testing-library/dom-testing-library/pull/616#issuecomment-643882204) with race conditions, so merger isn't happening right now 😞  I'm a bit torn on this because that would have meant that we could remove a ton of differences between Vue Testing Lib and React Testing Lib, but looks like we'll need to wait.

Closing this up (for now, hopefully!).

Maybe there's a way to handle `fireEvent` synchronously in Vueland, too…

Issue #83, opened 2019-08-13
Warn about v-model components not handled with fireEvent.change


afontcu:
More often than not people relies on `fireEvent.input` or `fireEvent.change` to handle v-model powered inputs, thus running into issues with input value not being properly updated.

I'm wondering if we should provide some kind of warning when that happens.

Does it make sense?

chiptus:
Do we have docs about it? I'm currently struggling with updating a custom component in a form using `fireEvent.update`

afontcu:
We do have a section in the [API docs](https://testing-library.com/docs/vue-testing-library/api#updateelem-value), and some examples in the repo: 

https://github.com/testing-library/vue-testing-library/blob/993c87f8742fcaaaa942c8ae04b63f503f4e5824/src/__tests__/select.js
https://github.com/testing-library/vue-testing-library/blob/993c87f8742fcaaaa942c8ae04b63f503f4e5824/src/__tests__/form.js

chiptus:
is `fireEvent.update` working with custom elements? like if I have a custom select box? in your examples you find the select component, but in the end, it's just a regular select. I'm trying to understand if all fireEvent.update expect is a component which can be used with v-model (because in my examples, it doesn't work)

afontcu:
If that's the case, would you mind opening up a new issue with a reproduction link? All `fireEvent.update` [does](https://github.com/testing-library/vue-testing-library/blob/master/src/vue-testing-library.js#L128) is guessing the input type it was called upon, and update its value accordingly.

Thanks!! 🙌

chiptus:
If I understand the code for [fireEvent](https://github.com/testing-library/dom-testing-library/blob/1233db0d343fa73fb2bc68fcbe816933adc2042e/src/events.js), this function just fires native events, right? if I want to update a component, I should update it as the user does, for example in my selector, I would click open, and choose the right value. right?

afontcu:
> this function just fires native events
Yes, it is aligned with the guiding principle of Testing Library: keeping things as close as how your software is used.

> if I want to update a component, I should update it as the user does, for example in my selector, I would click open, and choose the right value. right?
Exactly! Interact with your component as any user would (or, well, as similar as possible).

---

`fireEvent.update` goes the extra mile and helps you with common use cases for `v-model`,  but I'm not 100% positive it would handle complex scenarios with custom events and compound components. If you find yourself stuck because of that, please share an example so we can work on it :)

chiptus:
[...]

PR #160, opened 2020-10-27
Add warning message to fireEvent input and change when called directly #83

jhack32:
I've made an assumption here that whenever we get a second arg, it's coming from a direct call but would be happy to change it if it's an incorrect assumption. Let me know what y'all think!

### DIFF
--- a/src/__tests__/fire-event.js
+++ b/src/__tests__/fire-event.js
@@ -119,6 +119,13 @@ const eventTypes = [
     elementType: 'div',
   },
 ]
+beforeEach(() => {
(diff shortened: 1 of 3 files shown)


afontcu:
Hi, thanks for this!

I gave this some more thought and now I wonder what would happen for people using Vue Testing Lib + [user-event](https://github.com/testing-library/user-event)… 🤔

jhack32:
@afontcu I could add some test cases for [user-event](https://github.com/testing-library/user-event).

PR #165, opened 2020-11-01
Move types from DefinitelyTyped to this repository

afontcu:
Following the steps of [DTL](https://github.com/testing-library/dom-testing-library/pull/530), [RTL](https://github.com/testing-library/react-testing-library/pull/690), and [CTL](https://github.com/testing-library/cypress-testing-library/pull/148), this PR brings types home.

Any feedback by TS pro's is really appreciated 🙌 @testing-library/core-maintainers


timdeschryver (approved):
LGTM

types/index.d.ts
```diff
@@ -0,0 +1,74 @@
...
+  container?: HTMLElement
+  baseElement?: HTMLElement
+}
+
+export type ConfigurationCallback<V extends Vue> = (
+  localVue: typeof Vue,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  store: Store<any>,
+  router: Router,
+) => Partial<ThisTypedMountOptions<V>> | void
```
afontcu, line 48: This line productes the following warning with latest config:

```
void is only valid as a return type or generic type variable
@typescript-eslint/no-invalid-void-type
```

not entirely sure I should be worries about this?

.eslintrc.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
+  parserOptions: {
+    parser: '@typescript-eslint/parser',
```
kentcdodds, line 3: Support for this is now built-in to the latest version of kcd-scripts (actually, eslint-config-kentcdodds, but that's what kcd-scripts uses).

afontcu: That's what I assumed when I checked the changelogs, but sadly enough [eslint-plugin-vue requires](https://eslint.vuejs.org/user-guide/#how-to-use-a-custom-parser) us to use `parserOptions.parser` option instead of `parser`, because the latest needs to be used internally to parse Vue files.

SFC are cool but come with their own headaches 😅

PR #166, opened 2020-11-03
Warn when users are using fireEvent.input or .change directly

jhack32:


src/vue-testing-library.js
```diff
@@ -115,9 +115,16 @@ async function fireEvent(...args) {
...
   await waitFor(() => {})
 }
+const changeOrInputEventCalledDirectly = (eventValue, eventKey) =>
+  eventValue && (eventKey === 'change' || eventKey === 'input')
 
 Object.keys(dtlFireEvent).forEach(key => {
   fireEvent[key] = async (...args) => {
+    if (changeOrInputEventCalledDirectly(args[1], key)) {
+      console.warn(
+        'By using "change" or "input", there may be unexpected issues. Please use "update" for a better experience.',
```
afontcu, line 125: A small suggestion to (1) write down the used event key (so people have a better feedback on where to look for) and (2) to display "fireEvent.", so that people have the exact answer right there in the warning. Any other suggestion is welcome :)

```suggestion
        `Using "fireEvent.{$key} may lead to unexpected results. Please use fireEvent.update() instead.`,
```

afontcu (commented):
Looks great! Left a small note on the specifics of the warning message

src/vue-testing-library.js
jhack32: This makes sense, I'll make that change, thanks!

src/vue-testing-library.js
```diff
@@ -112,9 +112,16 @@ async function fireEvent(...args) {
...
   await waitFor(() => {})
 }
+const changeOrInputEventCalledDirectly = (eventValue, eventKey) =>
+  eventValue && (eventKey === 'change' || eventKey === 'input')
 
 Object.keys(dtlFireEvent).forEach(key => {
   fireEvent[key] = async (...args) => {
+    if (changeOrInputEventCalledDirectly(args[1], key)) {
+      console.warn(
+        `Using "fireEvent.{$key} may lead to unexpected results. Please use fireEvent.update() instead.`,
```
afontcu, line 122: it should actually be

```suggestion
        `Using "fireEvent.${key} may lead to unexpected results. Please use fireEvent.update() instead.`,
```

so that it renders "change" or "input" 😉

jhack32: Thanks for catching that!

afontcu (approved):
👍  neat!

PR #170, opened 2020-11-13
Fix updateProps Function

ITenthusiasm:
# The Problem
I just discovered a bug in one of my projects with the `updateProps` function returned from `render`. Whenever `updateProps` is called multiple times, several errors are logged in the console and Jest is prevented from finishing its test run. This is a huge bug for obvious reasons. Anyone can reproduce this error by grabbing the (currently) latest version of VTL and making multiple `await`ed calls to `updateProps` in a test for a Vue component.

The logged errors seem to be related to asynchronicity. I think that since the `Promise` returned from `updateProps` is different from the one returned from `waitFor`, everything gets thrown out of balance when the developer tries to run tests with multiple `updateProps` calls.

The fix I'm proposing simply returns the `Promise` obtained from `wrapper.setProps` directly. I applied this fix to my `node_modules`, and my tests behaved properly afterwards.

# Changes
* Changed updateProps to return the Promise created from wrapper.setProps.

### DIFF
--- a/src/__tests__/update-props.js
+++ b/src/__tests__/update-props.js
@@ -14,8 +14,10 @@ test('calling render with the same component but different props does not remoun
   expect(getByTestId('number-display')).toHaveTextContent('1')
 
   await updateProps({number: 2})
-
   expect(getByTestId('number-display')).toHaveTextContent('2')
+
+  await updateProps({number: 3})
+  expect(getByTestId('number-display')).toHaveTextContent('3')
 
   // Assert that, even after updating props, the component hasn't remounted,
   // meaning we are testing the same component instance we rendered initially.
--- a/src/vue-testing-library.js
+++ b/src/vue-testing-library.js
@@ -80,10 +80,7 @@ function render(
     isUnmounted: () => wrapper.vm._isDestroyed,
     html: () => wrapper.html(),
     emitted: () => wrapper.emitted(),
-    updateProps: _ => {
-      wrapper.setProps(_)
-      return waitFor(() => {})
-    },
+    updateProps: _ => wrapper.setProps(_),
     ...getQueriesForElement(baseElement),
   }
 }


afontcu:
Hi! Thank you very much for this 🙌  It would be great if you could add a test that would fail with the previous implementation and that passes with the proposed one :)

ITenthusiasm:
I know the code change, but I don't seem to be able to run a check locally. After running an `npm install` with the `next` branch, I haven't been able to run tests properly on `master`. Also, I noticed that for some reason `package-lock.json` files aren't generated when an install occurs.

ITenthusiasm:
@afontcu Done.

Regarding my previous comment, it seems that
1. Performing `npm install` on `master` (or a branch on top of it)
2. Then performing `npm install` on `next` (or a branch on top of it)

Will cause problems to happen -- at least on certain machines. And the problem is that future installs on the `master` branch (or "related" branches) will permanently be broken, causing attempted tests to end in failure frequently. The `next` branch still seems to be able to run its tests just fine.

I'm running Ubuntu 18.04. After installing on `master`, then `next`, I was unable to perform a proper install on `master` anymore. More specifically, after performing an install on the `master` branch, I would still be _unable_ to run any tests successfully. I can't tell if this is an issue with the `master` branch, the `next` branch, my machine, or some combination in between. Something could be wrong with the packages too? But it complicates contributing because switching between `next` and `master` branches requires me to use 2 different local versions of the forked repo. I think it's something worth investigating.

afontcu:
Hi!

> I know the code change, but I don't seem to be able to run a check locally. After running an npm install with the next branch, I haven't been able to run tests properly on master. 

`next` branch is using Vue 3 and I recently merged [some changes regarding types](https://github.com/testing-library/vue-testing-library/pull/165) that might break that branch. That's OK, I'll be fixing them soon.

> Also, I noticed that for some reason package-lock.json files aren't generated when an install occurs.

This is intentional, see https://github.com/testing-library/vue-testing-library/blob/master/.npmrc

ITenthusiasm:
That makes sense. Thanks!

afontcu:
Thank you! 🤗

PR #171, opened 2020-11-14
chore: switch to github actions

nickmccurdy:
Generated with `yo kcd-oss`

See https://github.com/testing-library/dom-testing-library/pull/809 for rationale and history


nickmccurdy:
I've also fixed the ESLint config. This is technically broken on master, but Travis succeeds anyway because it caches old dependencies, which Actions doesn't.

afontcu (approved):
Looks great 😍 I guess there's no way to test the `release` job until there's, well, a release, right? 😂

nickmccurdy:
A similar config has already been used with the other packages so I think it should be fine.

PR #172, opened 2020-11-14
Use `@testing-library/jest-dom` in README.md

vdsabev:
Hey folks! I'm just starting out with `@testing-library/vue` and since the README said `You may also be interested in installing jest-dom` I tried installing it, only to then find out it was moved to `@testing-library/jest-dom`, so I had to install the other one.

Perhaps this little change could save other people some time - I'd be happy if you considered accepting it.

Thanks for reading 😊


afontcu (approved):
Nice catch, thanks!

PR #173, opened 2020-11-15
Replace `setProps` with `rerender` Function

ITenthusiasm:
* Removed `setProps` function (and tests) to confirm with Testing
Library standards.
* Added `rerender` function (and tests) to make the API more familiar to
the Testing Library family.


ITenthusiasm:
@afontcu [Done](https://github.com/testing-library/vue-testing-library/pull/137#issuecomment-727162909)!

ITenthusiasm:
Noticed that `@testing-library/vue` was being used as an alias to import from the actual JavaScript file in the repository, so I updated my code to conform to the standard.

However, in the future, I think it would be worth simply pointing to the actual file. Doing so would remove the need for the special Jest configuration. More importantly, it would create less confusion for other contributors. Taking a quick look at RTL and STL, they seem to use the regular file name as well instead of using an alias. I'm not going to address that in this PR though. Just pointing something out so it's not forgotten.

afontcu:
> Noticed that @testing-library/vue was being used as an alias to import from the actual JavaScript file in the repository, so I updated my code to conform to the standard. However, in the future, I think it would be worth simply pointing to the actual file.

I did so to make it simpler for users to just "copy&paste" the examples into their codebase and get them working, but I'm totally open to suggestions :)

src/vue-testing-library.js
```diff
@@ -47,7 +47,7 @@ function render(
   //   additionalOptions = configurationCb(router)
   // }
 
-  const wrapper = mount(
+  let wrapper = mount(
```
afontcu, line 50: No need to make it mutable, right? Could we just create a local variable inside of rerender at leave this as a `const`?

src/vue-testing-library.js
```diff
@@ -73,7 +73,26 @@ function render(
...
+        merge({
+          attachTo: container,
+          global: {
+            plugins,
+          },
+          ...mountOptions,
+          props,
+          // ...additionalOptions,
+        }),
+      )
```
afontcu, line 91: Would it make sense to create come sort of a `mountComponent` function (or something with a better name 😇) to avoid repeating this code block? Both rendering methods must have the same keys, and having them in a single place looks like a good thing.

afontcu: Hm, it could even include the unwrap + add lines below 🤔

src/__tests__/rerender.js
```diff
@@ -0,0 +1,46 @@
+import '@testing-library/jest-dom'
+import {defineComponent, h, computed} from 'vue'
+import {render} from '@testing-library/vue'
+import NumberDisplay from './components/NumberDisplay'
+
+test('calling rerender remounts the component and updates the props', () => {
```
afontcu, line 6: I would keep the suggestion to avoid using this method if it's not strictly necessary.

```suggestion
// It'd probably be better if you test the component that's doing the rerendering
// to ensure that the rerendered component is being updated correctly.
// That said, if you'd prefer to, for example, update the props of a rendered
// component, this function can be used to do so. 
test('calling rerender remounts the component and updates the props', () => {
```

afontcu (commented):
Looks good! Left some comments here and there

src/__tests__/rerender.js
ITenthusiasm: Done! I put the comment above the `test` line to be more like the old file if that's okay.

src/vue-testing-library.js
ITenthusiasm: [React Testing Library](https://github.com/testing-library/react-testing-library/blob/master/src/pure.js#L54) has something like that. At a first glance I think it should be doable.

src/vue-testing-library.js
ITenthusiasm: Yeah see, that's the hard part. Because the properties in the returned object point to `wrapper`, that binding is actually what needs to be updated. If `wrapper` is `const`, then it can't be reassigned. So when `rerender` is called, the original wrapper is destroyed, a new component is mounted, and the test looks fine. But any consecutive times that `rerender` is called in the same test (which is a realistic scenario), the the code will be working with a reference to the oldest wrapper, not the new one. So nothing new will be destroyed, and the DOM will get polluted very quickly. Moreover, users will be unable to take advantage of all utility functions that use `wrapper` (for the aforementioned reason).

Theoretically, there _might_ be some things we could try to keep it `const`? But in the interest of keeping things simple, I think it's best to use `let`. But as long as we're careful about how we use `wrapper`, we should be okay.

I promise I did try using `const` before my first commit, though! 😅

src/vue-testing-library.js
ITenthusiasm: I added my first iteration of the function. Let me know what you think! The 2 original blocks of code were slightly different due to the overwrite of `props`, so I had to include that in the arrow function. But besides that, I think it's okay. If you're fine with it, you can resolve this piece. If not, you can let me know! :smile:

src/vue-testing-library.js
ITenthusiasm: Due to what I just described, I updated the first test to call `rerender` twice. This should help avoid unexpected bugs, like in the case of the `updateProps` test on `master`.

Since I didn't submit a code change for this one, I'll leave the resolution up to you. If you think of any simple workarounds let me know. I'll keep thinking while the PR is open.

ITenthusiasm:
> I did so to make it simpler for users to just "copy&paste" the examples into their codebase and get them working, but I'm totally open to suggestions :)

Ah, I see. That makes sense. I'll think about it a bit more before opening up a PR. Thanks!

ITenthusiasm:
Hey! I noticed that you added some reactions to my comments that seemed positive. Since the PR hasn't been merged yet and I haven't seen any new comments, I just wanted to double check and make sure whether you wanted me to add any more changes or not.

afontcu:
> Hey! I noticed that you added some reactions to my comments that seemed positive. Since the PR hasn't been merged yet and I haven't seen any new comments, I just wanted to double check and make sure whether you wanted me to add any more changes or not.

Hey! yeah, it looks good :) I'll be probably merging and releasing it soon. Thanks! 🚀

Issue #177, opened 2020-11-21
Can't use `fireEvent.update` on input file


JeromeDeLeon:
**Describe the bug** A clear and concise description of what the bug is.

Since the arrival of `fireEvent.update` and `warning` we are somehow forced to update the test to use it, but the problem is that it doesn't have support for input file making our tests fail when trying to use that new API but getting a warning when using `input` or `change` event type.

**To Reproduce** Steps to reproduce the behavior:

1. Create a component with an input file in it.
2. Test it with `fireEvent.update`

**Expected behavior**

I expected it to accept fileList or array of mock file like what we're using `{ target: { files: [mockFile] } }` or from what I found from this [PR](https://github.com/kentcdodds/react-testing-library-examples/pull/1/files)

**Screenshots**

**Related information:**

- `@testing-library/vue` version: 5.4.1
- `Vue` version: 2.6.11
- `node` version: 12.19.0
- `npm` (or `yarn`) version: 6.14.8

Relevant code or config (if any)

```javascript

```

**Additional context**

afontcu:
Hi! Can you provide a failing test?

afontcu:
Fixed in #179

PR #179, opened 2020-11-21
Add support for input file on `fireEvent.update`

JeromeDeLeon:
https://github.com/testing-library/vue-testing-library/issues/177


afontcu:
Hi! Can't fix it right now, but you can use [userEvent.upload()](https://github.com/testing-library/user-event#uploadelement-file--clickinit-changeinit-). It works fine with your example.

That being said, I think the problem is that we should avoid setting `value` attribute when type is `file` ([source](https://github.com/testing-library/vue-testing-library/blob/master/src/vue-testing-library.js#L159-L161)). With that, the following test passes (copied from the test you linked in the issue):

```js
test('fireEvent.update should not crash with input file', async () => {
  const {getByTestId} = render({
    render(h) {
      return h('input', {
        attrs: {
          type: 'file',
          'data-testid': 'test-update',
        },
      })
    },
  })

  const file = new File(['(⌐□_□)'], 'chucknorris.png', {
    type: 'image/png',
  })

  const inputEl = getByTestId('test-update')

  Object.defineProperty(inputEl, 'files', {
    value: [file],
  })

  await fireEvent.update(inputEl)

  expect(console.warn).not.toHaveBeenCalled()
})

```

JeromeDeLeon:
```js
if (['checkbox', 'radio'].includes(type)) {
   elem.checked = true
  return fireEvent.change(elem)
} else if (type === 'file') {
  return fireEvent.change(elem, { target: { files: value } })
} else {
  elem.value = value
  return fireEvent.input(elem)
}
```

Why can't it be like this? instead of doing manual population of files property every time we use input file?

JeromeDeLeon:
Yeah, it works fine with `upload` and works just fine even with using just `fireEvent.change` manually, the only reason I want it by default is to avoid the warning that's all.

afontcu:
> ```js
> if (['checkbox', 'radio'].includes(type)) {
>    elem.checked = true
>   return fireEvent.change(elem)
> } else if (type === 'file') {
>   Object.defineProperty(elem, 'files', { value })
>   return fireEvent.change(elem)
> } else {
>   elem.value = value
>   return fireEvent.input(elem)
> }
> ```
> 
> Why can't it be like this? instead of doing manual population of files property every time we use input file?

That would make sense. However I'm not sure if we should step into `userEvent` realm with `fireEvent`.  The whole reaason  `fireEvent.update` exists is to overcome some limitations when it comes to async DOM updates. Now, handling `<input type="file">` differenty is probably unexpected, and better handled by `userEvent`. 

I'm ok with adding the `else if` case to avoid the warning when adding files, though 👍

JeromeDeLeon:
Yeah, you're right with the purpose of `fireEvent.update`. It should only be used when we're using the `v-model` or `value/ (change/input)` case and for the case of the input file, we're not allowed to set up a value so the ability to turn off the warning is fine.

afontcu:
I'd definitely add the `else if`, and also a test example on how to input a file! Fancy to do so?

src/__tests__/fire-event.js
```diff
@@ -215,6 +227,31 @@ test('fireEvent.update does not trigger warning messages', async () => {
   expect(console.warn).not.toHaveBeenCalled()
 })
 
+test('fireEvent.update should not crash with input file', async () => {
```
JeromeDeLeon, line 230: Is this test example not enough? @afontcu

afontcu: It is! But as mentioned, I wouldn't make `fireEvent.update()` smarter when it comes to type="file" inputs. I would just trigger the proper event, and then add a test showcasing how to handle files as in React Testing Library (and probably also a comment suggesting userEvent to do so:


```js
test('fireEvent.update should not crash with input file', async () => {
  const {getByTestId} = render({
    template: `<input type="file" data-testid="test-update" />`
  })

  const file = new File(['(⌐□_□)'], 'chucknorris.png', { type: 'image/png' })

  const inputEl = getByTestId('test-update')

  Object.defineProperty(inputEl, 'files', {
    value: [file],
  })

  await fireEvent.update(inputEl)

  expect(console.warn).not.toHaveBeenCalled()
})
```

afontcu: I added a suggestion with the change I mention above 👍

src/vue-testing-library.js
```diff
@@ -155,6 +155,9 @@ fireEvent.update = (elem, value) => {
       if (['checkbox', 'radio'].includes(type)) {
         elem.checked = true
         return fireEvent.change(elem)
+      } else if (type === 'file') {
+        Object.defineProperty(elem, 'files', {value})
+        return fireEvent.change(elem)
```
afontcu, line 160:
```suggestion
      } else if (type === 'file') {
        return fireEvent.change(elem)
```

(also not 100% sure if we should trigger a change event over an input one? 🤔 haven't checked, so it might be okay)

JeromeDeLeon: Seems pretty okay to me since we know we can't use v-model to an input file and the only reasonable event to use in this case is change.

afontcu (approved):

afontcu:
nice! 🙌

PR #181, opened 2020-11-22
Add env to turn off fireEvent.update warning

JeromeDeLeon:


src/vue-testing-library.js
```diff
@@ -117,7 +117,7 @@ const changeOrInputEventCalledDirectly = (eventValue, eventKey) =>
 
 Object.keys(dtlFireEvent).forEach(key => {
   fireEvent[key] = async (...args) => {
-    if (changeOrInputEventCalledDirectly(args[1], key)) {
+    if (process.env.WARN_EVENT_UPDATE === 'true' && changeOrInputEventCalledDirectly(args[1], key)) {
```
afontcu, line 120: To make it simpler, what about doing it in negative?
If we call the variable `VTL_SKIP_WARN_EVENT_UPDATE` (to make it similar to the other env var) we can then skip setting the default value in line 191.

JeromeDeLeon: Oh yeah, sounds reasonable.

afontcu (approved):
great!

PR #182, opened 2020-11-23
test: remove await from userEvent usage

nickmccurdy:
`userEvent` is entirely synchronous (except for `type`'s `delay` option which isn't used), so we should use it without `await` and encourage users to do the same.

### DIFF
--- a/src/__tests__/user-event.js
+++ b/src/__tests__/user-event.js
@@ -1,5 +1,5 @@
 import '@testing-library/jest-dom'
-import {render} from '@testing-library/vue'
+import {render, waitFor} from '@testing-library/vue'
 import userEvent from '@testing-library/user-event'
 import Form from './components/Form'
 import Select from './components/Select'
@@ -12,7 +12,7 @@ afterEach(() => {
   console.warn.mockRestore()
 })
 
-test('User events in a form', () => {
+test('User events in a form', async () => {
   const fakeReview = {
     title: 'An Awesome Movie',
     review: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit.',
@@ -43,7 +43,7 @@ test('User events in a form', () => {
 
   userEvent.click(wonderfulRadioInput)
   expect(wonderfulRadioInput).toBeChecked()
-  expect(initialSelectedRating).not.toBeChecked()
+  await waitFor(() => expect(initialSelectedRating).not.toBeChecked())
 
   const recommendInput = getByLabelText(/Would you recommend this movie?/i)
   userEvent.click(recommendInput)
--- a/src/__tests__/user-event.js
+++ b/src/__tests__/user-event.js
@@ -12,7 +12,7 @@ afterEach(() => {
   console.warn.mockRestore()
 })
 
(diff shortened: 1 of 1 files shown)


afontcu:
The checkbox group test fails if called synchronously :(

nickmccurdy:
What should we be using, `Vue.nextTick()`, `flushPromises()`, or `waitFor()`?

afontcu:
> What should we be using, `Vue.nextTick()`, `flushPromises()`, or `waitFor()`?

To keep it relevant to Testing Library users… I'd say `waitFor`? All these options would do the trick, though.

src/__tests__/user-event.js
```diff
@@ -24,49 +24,49 @@ test('User events in a form', async () => {
...
   const initialSelectedRating = getByLabelText(/Awful/i)
   const wonderfulRadioInput = getByLabelText(/Wonderful/i)
   expect(initialSelectedRating).toBeChecked()
   expect(wonderfulRadioInput).not.toBeChecked()
 
-  await userEvent.click(wonderfulRadioInput)
+  userEvent.click(wonderfulRadioInput)
   expect(wonderfulRadioInput).toBeChecked()
-  expect(initialSelectedRating).not.toBeChecked()
+  await waitFor(() => expect(initialSelectedRating).not.toBeChecked())
```
afontcu, line 46: Not 100% sure it's better for users to learn when/why they should use `waitFor` instead of just awaiting userSelect? (I honestly don't know 😂)

nickmccurdy:
These build failures are unrelated, but I've fixed them in #188 so merging isn't blocked.

src/__tests__/user-event.js
nickmccurdy: I assume you mean userEvent, not userSelect. I don't think we should encourage users to await `click` because it's not async, and awaiting it implies that the test code will wait for Vue to render (not necessarily true as far as I understand, because Vue renders asyncronously). Awaiting anything before this line (even `await null`) will pass the test. So while `await userEvent.click(...)` is simpler, I don't think it's very correct.

afontcu: 👍

afontcu: deal 😉

nickmccurdy: Want to merge this as is, or do you prefer any of the alternatives?

afontcu: I guess it's fine as is 😃 I wonder if ppl will understand/know the difference between "test is failing because Vue hasn't updated the DOM yet" and "test is failing because my component is actually wrong", but that's something beyond our control (as long as Vue updates async)

afontcu (approved):

src/__tests__/user-event.js
nickmccurdy: If we make userEvent async for Vue users, this can be reverted, and "test is failing because Vue hasn't updated the DOM yet" should be more obvious when an `await` is missing (some linters can even warn you about not awaiting an async function).

PR #183, opened 2020-11-24
Make fireEvent.update work with lazy modifier

kinoli:
- switched to use change event if lazy modifier is found on element
- updated form unit test to test lazy input and textarea


afontcu (approved):
good catch!

afontcu:
This is gonna be interesting to implement in Vue 3 😅

PR #191, opened 2020-12-09
feat: Reenable Vue Router for Vue 3

afontcu:
[Vue Router](https://github.com/vuejs/vue-router-next) is out of beta and looks like we already provide basic support.

### DIFF
--- a/src/__tests__/vue-router.js
+++ b/src/__tests__/vue-router.js
@@ -24,7 +24,7 @@ test('full app rendering/navigating', async () => {
 
   await fireEvent.click(getByTestId('about-link'))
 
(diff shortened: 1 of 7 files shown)


src/__tests__/vue-router.js
```diff
@@ -1,38 +1,34 @@
...
-//   const {queryByTestId} = render(App, {routes})
-
-//   expect(queryByTestId('location-display')).toHaveTextContent('/')
-
-//   await fireEvent.click(queryByTestId('about-link'))
-
-//   expect(queryByTestId('location-display')).toHaveTextContent('/about')
-// })
-
-// test('setting initial route', () => {
```
afontcu, line 29: AFAIK this test is no longer doable

Issue #198, opened 2021-01-08
about rerender api design


cuixiaorui:
Hi, I'm a little confused when using the rerender API

I wanted to update props, but then I realized that in the next release, I needed to call rerender api

However, I looked at the implementation of rerender and saw that the component instance was re-destroyed and then re-created

But my test requirement is to simulate that the user only updated props

For example, I did this earlier by calling the wrapper. setProps API for vue-test-utils-next

Now the new rerender behaves substantially differently than setProps

The former is destroyed to re-create components based on the new props
The latter is to change the props on a component that has been created previously

I need to have a consistent API and setProps behavior

afontcu:
Hi! Thanks for your message. 

Yep, you're right, I overlooked this bit: https://github.com/testing-library/vue-testing-library/pull/173/files#diff-25bf86bba97788343267cf1a217702b4825b65b411e7697b0daa7576a5eb79ecR23-R25

`instance-id` should be `1`.


That being said:

> I need to have a consistent API and setProps behavior

Moving from Vue 2 support to Vue 3 might yield different behaviors, so consistency is not assured. I understand you want a consistent API, but here we're trying to find the middle ground between vue test utils and testing library family.

---

Also, if you need for some reason to update props, it'd probably be better if you test the component that's doing the updating to ensure that the props are being updated correctly.

afontcu:
Any help is appreciated, by the way!

cuixiaorui:
感谢你的回复哈

我在 master 版本上找到了一个 updateProps 的api

从代码意图上来讲，updateProps 是个很好的 api

当初是因为什么原因把它移除了呢？

rerender 一个合理的设计方案可能是

保留之前的 wrapper 实例对象，只做更新操作

当然，从代码实现上来讲，销毁之后再创建是使代码实现变的更简单了

你对 rereader 合理的行为怎么看？


Thanks for your reply

I found an API for updateProps on the master version

updateprops is a nice API from a code intent point of view

What was the reason for its removal?

A reasonable design scenario for rerender might be：

Keep the previous Wrapper instance and only update it

Of course, from a code implementation point of view, it's much easier to destroy and then create

What do you think of the reasonable behavior of rereader?

afontcu:
Yep. I think we should:

1) Go back to the original behavior of setProps/rerender.
2) Find out if there's a way to keep the previous wrapper instance and update it, so it does not only update props, but any other information we pass down to the component.

[...]

PR #199, opened 2021-01-08
fix(rerender): Do not mount component again on rerender

### LINKED ISSUE #197: Unexpected parameter format for rerender() - v6 (next)
Hi,
The rerender function's signature on v6 is: 
```ts
rerender(props: object): Promise<void>
```
But this is how it works:
```ts
rerender({ props: { error: false } })
```
I think the `props` key is redundant and misleading here and it needs to be something like:
```ts
rerender({ error: false })
```

afontcu:
Closes #197 and #198

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -12,7 +12,7 @@ import {
 const mountedWrappers = new Set()
 
 function render(
-  TestComponent,
+  Component,
   {
     store = null,
     routes = null,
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -72,12 +66,7 @@ function render(
     unmount: () => wrapper.unmount(),
     html: () => wrapper.html(),
     emitted: () => wrapper.emitted(),
-    rerender: ({props}) => {
-      wrapper.unmount()
-      mountedWrappers.delete(wrapper)
-
-      wrapper = mountComponent(TestComponent, props)
-    },
+    rerender: props => wrapper.setProps(props),
```
afontcu, line 69: This is mostly it. Other changes are because we don't need a `wrapper` variable (and a `mountComponent` function) anymore.

PR #203, opened 2021-01-14
replaced queryByTestId with getByTestId in the vue-router test

SangilYun:
First of all, thank you for all the test examples. It really helped me to set up the test environment on my project.

I found that this test uses `queryBy*` to query the element with `location-display` testid

As it's mentioned in this [link](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-query-variants-for-anything-except-checking-for-non-existence), using `getBy*` query might fit better for this test I think.


afontcu (approved):
Nice catch! Thank you :)

PR #206, opened 2021-02-10
feat: Remove Router/Vuex for Vue 3

### LINKED ISSUE #195: VTL for Vue 3: The routes option does not work correctly
**Describe the bug**
When using the routes option of the render function, it throws up some warnings that have to do with the _history property and the asynchronic handling that vue router 4 has. This causes you to not be able to use the routes option correctly.

**To Reproduce**
```typescript
import { render, screen } from '@testing-library/vue'
import '@testing-library/jest-dom'
import { defineComponent } from 'vue'
import { RouteRecordRaw } from 'vue-router'

test('Component with route', () => {
  const ComponentA = defineComponent({
    name: 'ComponentA',
    props: {
      to: {
        type: Object,
        required: true
      }
    },
    template: `<router-link :to="to" role="button">Learn More</router-link>`
  })

  const ComponentB = defineComponent({
    name: 'ComponentB'
  })

  const routeRecordRaw: RouteRecordRaw = {
    path: '/',
    name: 'componentB',
    component: ComponentB
  }

  const to = { name: routeRecordRaw.name }

  render(ComponentA, {
    props: { to },
    routes: [routeRecordRaw]
  })

  const button = screen.getByRole('button')

  expect(button).toBeInTheDocument()
  expect(button?.getAttribute('href')).toBe(routeRecordRaw.path)
})
```
This throws the following warning
```bash
console.warn node_modules/vue-router/dist/vue-router.cjs.js:75
    [Vue Router warn]: Unexpected error when starting the router: TypeError: Cannot read property '_history' of null
```
The warning is known. This is explained in the vue-test-utils v2 documentation
https://vue-test-utils.vuejs.org/v2/guide/vue-router.html#with-a-real-router

So to solve it apply what is indicated in documentation
```typescript
import { render, screen } from '@testing-library/vue'
import '@testing-library/jest-dom'
import { defineComponent } from 'vue'
import { createRouter, createWebHistory, RouteRecordRaw } from 'vue-router'

test('Component with route', async () => {
  const ComponentA = defineComponent({
    name: 'ComponentA',
    props: {
      to: {
        type: Object,
        required: true
      }
    },
    template: `<router-link :to="to" role="button">Learn More</router-link>`
  })

  const ComponentB = defineComponent({
    name: 'ComponentB'
  })

  const routeRecordRaw: RouteRecordRaw = {
    path: '/',
    name: 'componentB',
    component: ComponentB
  }

  const router = createRouter({
    history: createWebHistory(process.env.BASE_URL),
    routes: [routeRecordRaw]
  })

  router.push('/')
  await router.isReady()

  const to = { name: routeRecordRaw.name }

  render(ComponentA, {
    props: { to },
    global: {
      plugins: [router]
    }
  })

  const button = screen.getByRole('button')

  expect(button).toBeInTheDocument()
  expect(button?.getAttribute('href')).toBe(routeRecordRaw.path)
})

```
**Expected behavior**
Be able to correctly use the routes option provided by the render function

**Related information:**

- `@testing-library/vue` version: ^6.3.1
- `Vue` version: ^3.0.0
- `node` version: 14.4.0
- `npm` version: 6.14.9

### LINKED ISSUE #205: [RFC] Remove custom router/vuex management for Vue 3
* Target Major Version: 6.x (VTL version for Vue 3)

# Summary 

Remove custom integrations for Vue Router and Vuex, and leave it to users to configure them as they'd do with any other plugin.

Currently, VTL for Vue 3 offers basic support for Router and Vuex, similar to VTL for Vue 2. It requires router/vuex library if the appropriate render option is provided, creates the plugin and adds it to `global.plugins`. See [source](https://github.com/testing-library/vue-testing-library/blob/next/src/index.js#L29-L41).

# Basic example

VTL would remove support for `store` and `routes` keys on render:

```diff
render(Component, {
  props: { ... },
-  store: {},
-  routes: {}
})
```

in favor of VTU' `global.plugins` ([docs](https://next.vue-test-utils.vuejs.org/api/#global-plugins)).

```js
import { render } from '@testing-library/vue'
import Component from './Component'
import myStore from './store'
import myRoutes from './routes'
import { createRouter, createWebHistory } from 'vue-router'

const store = createStore(myStore)

// Assume myRoutes contains an array with routes.
const router = createRouter({
  history: createWebHistory(),
  myRoutes
})

test('test with router and vuex', () => {
  render(Component, {
    global: {
      plugins: [store, router]
    }
  })  
})
```

This is the suggested way to install Router/Vuex in VTU 2.x, too (see [here](https://next.vue-test-utils.vuejs.org/guide/advanced/vue-router.html#using-a-real-router) and [here](https://next.vue-test-utils.vuejs.org/guide/advanced/vuex.html#testing-with-a-real-vuex-store)).

Strictly speaking, removing integration with Vuex is not needed. However, it would be weird to only support one of the official Vue plugins, and treat the other one as a regular plugin.

# Motivation

This RFC is motivated by the discussion in https://github.com/testing-library/vue-testing-library/issues/195. In short:

Vue Router is now async, so it needs special treatment when it comes to navigating (simple – just use `waitFor` or `findByXXX` queries), but also if user needs to assert something related to first navigation (Router exposes a `router.isReady()` helper). This makes an out-of-the-box management from VTL a bit… weird.

# Detailed design

There's not much to do on VTL. VTU 2.x offers a comfortable way to install global plugins to vue 3 app through `global.plugins`, so this is way simpler than Vue 2 (where we needed to use a local Vue instance).

We're already installing anything provided through `globals.plugins`,  so we should good to go. It's just a matter of docs + removing `routes`/`store` options.

The change would make users responsible for setting up router/vuex properly. This is the exact same thing React Testing Library users have to do (see an [example for React Router](https://testing-library.com/docs/example-react-router) and another one for [Reach Router](https://testing-library.com/docs/example-reach-router)).

# Drawbacks

* Users need to initialize router/vuex. This is not the case with VTL for Vue 2, where we import the libraries when needed and attach them to the local vue instance.

# Alternatives

* Keep handling Router/Vuex the best way we can. In the original issue I outlined some alternatives that involved making `render` async or exposing a `isRouterReady` helper method. Both of the alternatives help us hide some of the complexity, but not all of it.

# Adoption strategy

We're still in beta here, so any breaking change is undesirable but expected. 

People would need to:

1) Remove `routes` and `store` keys from their tests.
2) Wrap `routes` with `VueRouter.createRouter` (and potentially register a `history` method.
3) Wrap `store` with `Vuex.createStore`.
4) Move both variables down to `global.plugins`.

We could display a warning message to make people aware of why router/vuex tests start failing: 

```js
function render(
  Component,
  {
    store = null,
    routes = null,
    // ...
  },
) {
  if (store || routes) {
    console.warning(`Using store/routes options is now deprecated.
      Please provide them as plugins through 'global.plugins' parameter.
      See here an example of how to integrating them into VTL: (add link)`)
  }
}
```

The message would be removed before marking VTL for Vue 3 as stable.

afontcu:
Closes #205. Closes #195.


types/index.d.ts
```diff
@@ -33,8 +31,14 @@ type VueTestUtilsRenderOptions = Omit<
...
-  store?: StoreOptions<{}>
-  routes?: RouteRecordRaw[]
+  /**
+   * @deprecated Use `global.plugins` array instead.
+   */
+  store: Record<string, any>
+  /**
+   * @deprecated Use `global.plugins` array instead.
+   */
+  routes?: Record<string, any>
```
ITenthusiasm, line 41: I think the `@deprecated` notice is good. The types seem a little weird though -- particularly `routes`.  It's technically not the real type. The imports could be kept to ensure the correct types appear... But at the same time it doesn't matter too much since it can't be used? What are your thoughts?

types/test.ts
```diff
@@ -77,12 +77,8 @@ export function testOptions() {
     },
     global: {
       config: {isCustomElement: _ => true},
+      plugins: [],
     },
-    store: {
-      state: {count: 3},
-      strict: true,
-    },
-    routes: [{path: '/', component: () => SomeComponent, name: 'route name'}],
```
ITenthusiasm, line 85: Is there a reason you removed the `store` and the `state` entirely instead of putting it in `global.plugins`?

src/index.js
```diff
@@ -24,26 +24,15 @@ function render(
...
-    const requiredRouter = require('vue-router')
-    const {createRouter, createWebHistory} =
-      requiredRouter.default || requiredRouter
-
-    const routerPlugin = createRouter({history: createWebHistory(), routes})
-    plugins.push(routerPlugin)
+  if (store || routes) {
+    console.warn(`Providing 'store' or 'routes' options is now deprecated.
+You need to create a router/vuex plugin and provide it through 'global.plugins'.
+Check out the test examples on GitHub for further details.`)
```
ITenthusiasm, line 30: > You need to create a router/vuex plugin

Is it technically considered a plugin or an instance? (Don't fully know the official lingo.)

Also, any thoughts about adding a link to the file? I think it would help people find the test example a lot more easily. I remember having a hard time finding VTL examples when I first tried it out. (Speaking of which, I still need to get to that docs thing. :eyes:) Hopefully it wouldn't make the logs too fat?

src/index.js
```diff
@@ -24,26 +24,15 @@ function render(
...
   }
 
   const wrapper = mount(Component, {
     ...mountOptions,
     attachTo: container,
-    global: {...mountOptions.global, plugins},
```
ITenthusiasm, line 46: Really cool to see all the lines that get saved just from this one change. :smile:

src/__tests__/vuex.js
```diff
@@ -26,7 +32,7 @@ test('can render with vuex with defaults', async () => {
 
 test('can render with vuex with custom initial state', async () => {
   const {getByTestId, getByText} = renderVuexTestComponent({
-    state: {count: 3},
+    state: () => ({count: 3}),
```
ITenthusiasm, line 35: So `state` accepts a function similar to `data` now? (Please excuse the noob questions. :sweat_smile:)

src/__tests__/vue-router.js
```diff
@@ -1,26 +1,38 @@
...
+  const {findByText, getByText, getByTestId} = render(App, {
+    global: {
+      plugins: [router],
+    },
+  })
 
   // Vue Router navigation is async, so we need to wait until the
   // initial render happens
   expect(await findByText('You are home')).toBeInTheDocument()
+  expect(getByTestId('location-display')).toHaveTextContent('/')
```
ITenthusiasm, line 35: Out of curiosity, any reason for checking the location display instead of checking the path on `window.location.href`?

ITenthusiasm (commented):
Although I've contributed a little, I haven't seen the entire codebase. So some curious questions will pop up here and there. :sweat_smile:

types/index.d.ts
afontcu: Actually I was tempted to use `any`, as it won't really work and user should be replacing this. Not entirely sure why I ended up setting `Record<string, any>`, looks like a leftover 😂 I think I'll leave it as `any` because, as you mention, they can't be used and are only defined here so they can be marked as `@deprecated`. Thanks!

src/__tests__/vuex.js
afontcu: This is actually not related to the change of this PR, so I'll revert to prevent any confusion.

src/__tests__/vue-router.js
afontcu: Hmm not really, we simply added an element that renders the route to the DOM so we don't suggest checking against "implementation details" such as window.location.href ;)

afontcu: > Although I've contributed a little, I haven't seen the entire codebase. So some curious questions will pop up here and there. 😅

Thank you for your thoughts!

afontcu:
> Although I've contributed a little, I haven't seen the entire codebase. So some curious questions will pop up here and there. 😅

Thank you for your thoughts!

src/index.js
afontcu: > Is it technically considered a plugin or an instance?

Yeah, I think they should be called instances, good point. Fixing it.

> Also, any thoughts about adding a link to the file?

Thought about it but felt a bit "too much". Once docs thing is available at testing-library.com, we could then set a short link alias so the output isn't that verbose 👍

types/test.ts
afontcu: this file just serves as a test for types, so just making sure that `global.plugins` exists would be enough

ITenthusiasm: Ahhhhh, I see. Yeah that makes sense.

src/index.js
ITenthusiasm: Yeah, that's fair. :+1:

src/__tests__/vue-router.js
ITenthusiasm: Fair enough :smirk:

codfish (approved):

codfish:
hey @afontcu! looking forward to this change! Out of curiosity, when do you see this being released? I'm migrating to VTL now, but wondering if I should continue on my path or wait a little bit for this change to be released. Also, should we expect this to result in v6?

afontcu:
> hey @afontcu! looking forward to this change! Out of curiosity, when do you see this being released? I'm migrating to VTL now, but wondering if I should continue on my path or wait a little bit for this change to be released. Also, should we expect this to result in v6?

Hi! Yeah, I plan on releasing it next week. And yeah, I'll release it as a minor version of 6.x, the major range version for Vue 3 :)

src/index.js
```diff
@@ -24,26 +24,15 @@ function render(
...
-    const routerPlugin = createRouter({history: createWebHistory(), routes})
-    plugins.push(routerPlugin)
+  if (store || routes) {
+    console.warn(`Providing 'store' or 'routes' options is no longer available.
+You need to create a router/vuex instance and provide it through 'global.plugins'.
+Check out the test examples on GitHub for further details.`)
   }
 
   const wrapper = mount(Component, {
     ...mountOptions,
```
codfish, line 34: https://testing-library.com/docs/vue-testing-library/api/#options

in the docs it says `props` is supported, but it doesn't look like it here, but maybe I'm missing something? Also doesn't seem like `baseElement` and `container` are _not_ shown in the docs?

afontcu: Hi! 

> in the docs it says props is supported, but it doesn't look like it here, but maybe I'm missing something?
Docs are still meant to match VTL for Vue 2. This is why it explicitly says that `props` is added. In Vue 3, Vue Test 

Utils already accepts a `prop` key so there's nothing to do on our side.

> Also doesn't seem like baseElement and container are not shown in the docs?

Yep, that's true. Fancy to open up a PR? 😉  We could use [RTL docs](https://testing-library.com/docs/react-testing-library/api#render-options) as example.

_btw, if you have further questions regarding VTL for Vue 3, feel free to open up new issues so this PR keeps focused on router/vuex :)_

codfish: @afontcu https://github.com/testing-library/testing-library-docs/pull/767 done!

PR #209, opened 2021-02-17
chore: Cleanup structure

afontcu:
- [x] Split large main file into smaller, focused ones.
- [x] Bump deps.


src/fire-event.js
```diff
@@ -0,0 +1,93 @@
+/* eslint-disable testing-library/no-wait-for-empty-callback */
+import {waitFor, fireEvent as dtlFireEvent} from '@testing-library/dom'
+
+// Vue Testing Lib's version of fireEvent will call DOM Testing Lib's
+// version of fireEvent. The reason is because we need to wait another
+// event loop tick to allow Vue to flush and update the DOM
+// More info: https://vuejs.org/v2/guide/reactivity.html#Async-Update-Queue
+
+async function fireEvent(...args) {
```
codfish, line 9: question out of curiosity. when using VTL, should we always use the `fireEvent` exported by this library or instead should we use `user-event`? with RTL, user-event is suggested https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-testing-libraryuser-event

afontcu: hi! Using user-event is advisable, as it is closer to the user. We even showcase some examples: https://github.com/testing-library/vue-testing-library/blob/master/src/__tests__/user-event.js

PR #214, opened 2021-03-31
chore(lint): disable unbound-method rule

MatanBobi:
This rule is currently breaking for some reason in our tests:
https://github.com/testing-library/vue-testing-library/runs/2226695857?check_suite_focus=true


afontcu:
Thank you! 😍

PR #232, opened 2021-05-17
feat: allow passing instantiated Vuex store

blimmer:
As currently written, the `store` property passed to the `render` function must be `StoreOptions`, which is a plain Javascript object passed to `new Vuex.Store()` in the render function.

This is great for simple components and should definitely still be a supported option. However, it would be really useful for us to be able to pass a complex, instantiated Vuex store (generated by our common test helpers) as well.

This PR accomplishes this behavior by checking if the passed `store` is an `instance of Vuex.Store` and skipping the `new Vuex.Store()` call in that case.

If this seems reasonable, I can also update the API documentation in the other repository 😄 . Thanks for your time & consideration!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -40,6 +40,7 @@ export interface RenderOptions<V extends Vue, S = {}>
   // The props and store options special-cased by Vue Testing Library and NOT passed to mount().
   extends Omit<ThisTypedMountOptions<V>, 'store' | 'props'> {
   props?: object
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
   store?: StoreOptions<S>
   routes?: RouteConfig[]
   container?: Element
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -40,7 +40,6 @@ export interface RenderOptions<V extends Vue, S = {}>
   // The props and store options special-cased by Vue Testing Library and NOT passed to mount().
   extends Omit<ThisTypedMountOptions<V>, 'store' | 'props'> {
   props?: object
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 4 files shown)


afontcu (approved):
LGTM! Sorry for the late response, busy weeks. Merging this one once CI passes :)

afontcu:
Looks like there are some typing issues – probably due to the extended type for the `store` option? https://github.com/testing-library/vue-testing-library/pull/232/checks?check_run_id=2820969444#step:5:100

blimmer:
Thanks @afontcu - I updated this PR and [tested it on my fork](https://github.com/blimmer/vue-testing-library/runs/2834246232?check_suite_focus=true). I'm a little surprised that `StoreOptions` and `Store` are interchangeable but, if the test suite is working properly, it looks like they are!

types/test.ts
```diff
@@ -127,6 +128,30 @@ export function testConfigCallback() {
   })
 }
 
+export function testInstantiatedStore() {
+  render(SomeComponent, {
+    store: new Vuex.Store({
```
blimmer, line 133: I added this test to ensure that a `Store` could be provided as `StoreOptions<S>` and it looks like it works!

types/index.d.ts
```diff
@@ -40,6 +40,7 @@ export interface RenderOptions<V extends Vue, S = {}>
   // The props and store options special-cased by Vue Testing Library and NOT passed to mount().
   extends Omit<ThisTypedMountOptions<V>, 'store' | 'props'> {
   props?: object
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
afontcu, line 43:
```suggestion
```

afontcu:
Thank you for this! 🎉

PR #238, opened 2021-06-23
fix: use Vue Router in abstract mode

### LINKED ISSUE #210: Broken test isolation of VueRouter instances

`VueRouter` instances are not erased properly after completing a test case. 

```sh
git clone git@github.com:maxarndt/vue-testing-library.git
npm i
npm test
```

**Expected behavior**

All tests should pass.

**Screenshots**

```js
 FAIL  src/__tests__/vue-router.js
  ✓ full app rendering/navigating (51 ms)
  ✕ another app rendering/navigating (13 ms)
  ✓ setting initial route (6 ms)

  ● another app rendering/navigating

    TestingLibraryElementError: Unable to find an element with the text: /you are home/i. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.

    <body>
      <div>
        <div>
          <a
            class="router-link-active"
            data-testid="home-link"
            href="#/"
          >

        Home

          </a>

          <a
            aria-current="page"
            class="router-link-exact-active router-link-active"
            data-testid="about-link"
            href="#/about"
          >

        About

          </a>

          <div>
            You are on the about page
          </div>

          <div
            data-testid="location-display"
          >
            /about
          </div>
        </div>
      </div>
    </body>

      29 |   const {getByTestId, getByText} = render(App, {routes})
      30 |
    > 31 |   expect(getByText(/you are home/i)).toBeInTheDocument()
         |          ^
      32 |   expect(getByTestId('location-display')).toHaveTextContent('/')
      33 |
      34 |   await fireEvent.click(getByTestId('about-link'))

      at Object.getElementError (node_modules/@testing-library/dom/dist/config.js:37:19)
      at node_modules/@testing-library/dom/dist/query-helpers.js:90:38
      at node_modules/@testing-library/dom/dist/query-helpers.js:62:17
      at getByText (node_modules/@testing-library/dom/dist/query-helpers.js:111:19)
      at _callee2$ (src/__tests__/vue-router.js:31:10)
      at tryCatch (node_modules/regenerator-runtime/runtime.js:63:40)
      at Generator.invoke [as _invoke] (node_modules/regenerator-runtime/runtime.js:293:22)
      at Generator.next (node_modules/regenerator-runtime/runtime.js:118:21)
      at asyncGeneratorStep (node_modules/@babel/runtime/helpers/asyncToGenerator.js:3:24)
      at _next (node_modules/@babel/runtime/helpers/asyncToGenerator.js:25:9)
      at node_modules/@babel/runtime/helpers/asyncToGenerator.js:32:7
      at Object.<anonymous> (node_modules/@babel/runtime/helpers/asyncToGenerator.js:21:12)
```

**Related information:**

- `@testing-library/vue` version: HEAD at 7bef579
- `Vue` version: `"vue": "^2.6.12"`
- `node` version: `v12.18.4`
- `npm` version: `6.14.9`

**Additional context**

I just added two assertions to an already existent test case and duplicated the test to be executed twice in a row (see https://github.com/maxarndt/vue-testing-library/commit/306550fa8557c43adcdc5cf763143430b791ad30). 

Both tests should pass if the testing environment is cleaned up properly.
Unfortunately this is not the case.

blimmer:
This PR uses `mode: 'abstract'` when creating a Vue Router instance.

By default, the Vue Router uses 'hash' mode ([docs](https://router.vuejs.org/api/#mode)) which, in test environments with JSDOM, updates the `window.location.href` hash value. In jest, the JSDOM environment is not reset between tests, which means that state appears to leak between tests, as mentioned in #210.

I did some digging through the vue-test-utils issue tracker and found https://github.com/vuejs/vue-test-utils/issues/1681, which talks about this issue in more depth. The recommended solution from that thread is to use `abstract` mode, which guarantees you start with a clean Vue Router each time it's instantiated.

Alternately, if this solution does not seem acceptable, users can utilize a lifecycle hook, such as Jest's `beforeEach` to reset the state between tests. For example:

```ts
beforeEach(() => {
  window.location.replace('http://localhost/');
});
```

Fixes #210

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -39,7 +39,8 @@ function render(
     const VueRouter = requiredRouter.default || requiredRouter
     localVue.use(VueRouter)
 
-    router = new VueRouter({routes})
+    router = new VueRouter({routes, mode: 'abstract'})
+    router.push('/')
   }
 
   if (configurationCb && typeof configurationCb === 'function') {
--- a/src/__tests__/vue-router.js
+++ b/src/__tests__/vue-router.js
@@ -32,3 +32,9 @@ test('setting initial route', () => {
 
   expect(getByTestId('location-display')).toHaveTextContent('/about')
 })
+
+test('router state is not shared between tests', () => {
+  const {getByTestId} = render(App, {routes})
+
+  expect(getByTestId('location-display')).toHaveTextContent('/')
+})


src/render.js
```diff
@@ -39,7 +39,8 @@ function render(
     const VueRouter = requiredRouter.default || requiredRouter
     localVue.use(VueRouter)
 
-    router = new VueRouter({routes})
+    router = new VueRouter({routes, mode: 'abstract'})
+    router.push('/')
```
blimmer, line 43: This felt like the most reasonable default to use, since in `abstract` mode the stack is empty. In the default `hash` mode with a clean JSDOM environment, this would also be the behavior.

afontcu: Hi! Let me take a closer look, as even though this fixes the issue, we might have some unexpected behavior here for users who would expect a regular, history-based router (not entirely sure what are the differences with `abstract` mode).

Maybe we could follow a similar approach to https://github.com/testing-library/vue-testing-library/pull/232, where we allow users to pass their own instance of Vuex. This would allow people to set their own router mode while also preserving backwards compatibility and consistency between vuex/router options. What do you think about it?

blimmer: It's a fair point that users might expect a hash-based / history router to be used with this library. In my case, at least, having to pass the entire Vue Router would be a bit of a pain to have to do with each test, as `abstract` mode is the configuration I'd prefer to use throughout all my tests. So, at least initially, my thought is that some kind of global config file would be useful, for example:

```js
// vue-testing-library.config.js
module.exports = {
  vueRouterMode: 'abstract',
};
```

However, it looks like other `testing-library` plugins don't have a global configuration file where a user could set something like this. For instance, `react-testing-library` has support for a custom `render` function ([docs](https://testing-library.com/docs/react-testing-library/setup/)) where the user can set defaults.

So perhaps the way forward is to allow passing an instantiated Vue Router, but then also provide docs for a custom `render` function like the `react-testing-library` docs above? Either way, I think it also makes sense to add a point to the "FAQs" page about this, as it's a bit confusing no matter what we decide to do.

I'm happy to tackle these updates depending on the maintainer's preferences. Thanks!

afontcu: > For instance, react-testing-library has support for a custom render function (docs) where the user can set defaults

We can do the same thing here at Vue Testing Lib:

```js
import VueRouter from 'vue-router'
import MyComponent from './MyComponent'
import routes from './routes'

function customRender() {
  const router = new VueRouter({ routes, mode: 'abstract' })
  router.push('/')

  return render(MyComponent, { routes: router })
}

// and then

test('foo', () => {
  const { getByText } = customRender()
})
```

This would allow people to set up their custom Router's mode a single time

blimmer:
Closing in favor of #239

PR #241, opened 2021-06-29
fix: support TypeScript v3

### LINKED ISSUE #240: Type errors with TypeScript 3.9.7
**Describe the bug** `tsc` compile fails with Typescript v3.9.7

See https://github.com/blimmer/vue-testing-library-issue / https://github.com/blimmer/vue-testing-library-issue/runs/2946243182?check_suite_focus=true

**Expected behavior**

The Typescript compile should pass with TypeScript v3.

**Screenshots**

Screen Shot 2021-06-29 at 2 31 48 PM

**Related information:**

- `@testing-library/vue` version: 5.8.0
- `Vue` version: N/A
- `node` version: 14
- `npm` (or `yarn`) version: N/A

```javascript
import '@testing-library/vue'
```

**Additional context**

Shouldn't the TypeScript check step of the CI system catch this?

blimmer:
I'm not sure if there's a more clever way to do this, but this should work. The linked issue has a reproduction example.

Fixes #240

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -46,16 +46,15 @@ export interface RenderOptions<V extends Vue, S = {}>
   baseElement?: Element
 }
 
-type ConfigurationArgs = [
-  localVue: typeof Vue,
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  store: Store<any>,
-  router: Router,
-]
-
 export type ConfigurationCallback<V extends Vue> =
-  | ((...args: ConfigurationArgs) => Partial<ThisTypedMountOptions<V>>)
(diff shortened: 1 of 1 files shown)


types/index.d.ts
```diff
@@ -46,16 +46,15 @@ export interface RenderOptions<V extends Vue, S = {}>
   baseElement?: Element
 }
 
-type ConfigurationArgs = [
```
afontcu, line 49: out of curiosity, is this just a syntax change because TS 3.x does not support this type definition syntax?

blimmer: Correct - I believe that format is called "named tuple" which is a new TS v4 feature.

PR #244, opened 2021-07-13
chore: upgrade to kcd-scripts 10

nickmccurdy:
Rather than upgrade to kcd-scripts 11 and use Jest 27 which will require some updates to vue-jest, we can just upgrade to kcd-scripts 10 for now.


afontcu:
good point! merging 👌

Issue #187, opened 2020-11-28
Vue testing library depends on vuex and vue-router types


sapagat:
Hi there! For my surprise importing the `@testing-library/vue` module raises a compilation error due to a dependency with `vue-router` and `vuex`.

```
 ERROR  Failed to compile with 2 errors

These dependencies were not found:

* vue-router in ./node_modules/@testing-library/vue/dist/vue-testing-library.js
* vuex in ./node_modules/@testing-library/vue/dist/vue-testing-library.js

To install them, you can run: npm install --save vue-router vuex
  [=========================] 100% (completed)

 WEBPACK  Failed to compile with 2 error(s)

Error in ./node_modules/@testing-library/vue/dist/vue-testing-library.js

  Module not found: 'vue-router' in '/home/sam/sandbox/opencocos/app/node_modules/@testing-library/vue/dist'

Error in ./node_modules/@testing-library/vue/dist/vue-testing-library.js

  Module not found: 'vuex' in '/home/sam/sandbox/opencocos/app/node_modules/@testing-library/vue/dist'
```

I've tried installing both dependencies and test pass, but I believe it shouldn't be mandatory in order to run tests, right?

Notice: I'm not using Jest. I'm running tests with Mocha.

afontcu:
Hi! Yes it does, because VTL accepts a `routes` and `store` objects, which autoloads Router and Vuex: https://github.com/testing-library/vue-testing-library/blob/master/src/vue-testing-library.js#L36-L49

I once thought that they should be listed as peerDeps, but since they are only required if these keys are there…

sapagat:
I've double checked and no `routes` nor `store` are passed to the render method. It may be a webpack issue that detects the packages being required?

I've bypassed the issue by requiring both packages in order to use VTL. I believe it should be optional, and currently it isn't. Don't know how to solve it... =(

afontcu:
Hm, can you share the test that makes both deps fail?

kauanschaeffer:
Having similtar issue, with vue-router only (the project doesnt have a dependency o vue-router): 

```
6:35 Cannot find module 'vue-router' or its corresponding type declarations.
    4 | import {ThisTypedMountOptions, VueClass} from '@vue/test-utils'
    5 | import {Store, StoreOptions} from 'vuex'
  > 6 | import Router, {RouteConfig} from 'vue-router'
      |                                   ^
    7 | // eslint-disable-next-line import/no-extraneous-dependencies
    8 | import {
[...]

PR #246, opened 2021-07-30
fix(types): Remove Router and Vuex dependency from types

afontcu:
Closes https://github.com/testing-library/vue-testing-library/issues/187

I'm not entirely sure this is the best way to fix the issue, but I want to get the ball rolling.


afontcu:
releasing this one. As stated before, if anyone comes up with a better way of handling Router/Vuex types without actually importing them, I'm all ears :)

Issue #92, opened 2019-08-26
How to use with Nuxt.js


rubenmoya:
Hello!

I'm trying to use `vue-testing-library` with Nuxt.js and I'm facing a problem with `asyncData` not being called.

I've been reading for a while, and some people call `wrapper.vm.$options.asyncData()` manually, but that feels super dirty to me, also I don't think `vm` is available in `vue-testing-library`

Any idea of how can I do it? I ran out of ideas to be honest.

I've seen @afontcu post about [testing api calls](https://afontcu.dev/testing-api-calls/) but it uses the `created` life cycle method

Thank you!

afontcu:
Hi Rubén! This is quite a good question 🤔 

I can't think of a way to test a page with `asyncData` (or `fetch`), since they are not part of the usual component lifecycle. AFAIK same issue should happen in Vue Test Utils.

Not sure there's a way for us to fix this, to be honest. You could try and use tools such as `nuxt.renderRoute` as described [here](https://nuxtjs.org/guide/development-tools#end-to-end-testing).

If you come up with any good idea, please do share it! 🙌

rubenmoya:
To be honest I didn't came up with a good solution, I just went ahead and used Cypress 🤷‍♂

afontcu:
Thanks for the update! Looks like the way I would've followed, too :)

Martin7mind:
i did this and it kind of works.. but is there a better solution now? 

```
export const mountWithAsyncData = async (component, options) => {
  let data
  if (component.asyncData) {
    const asyncDataReq = component.asyncData({ app: { $axios: axios } })
    await options.respondToAsyncData()
    data = await asyncDataReq
  }

  return mount(component, {
    data: () => data,
    ...options
  })
}
```

damafeez:
This is what I came up with, I might end up converting it into a tiny package later.
```
import { mount } from '@vue/test-utils'

export const mountWithAsyncData = async (
  component,
  { asyncDataContext = {}, ...options } = {},
) => {
  const returnedData = options.data === 'function' ? options.data() : null
  if ('data' in options && returnedData === null)
    throw new Error('data should be a function that returns an object literal')
  if (typeof component.asyncData === 'function') {
    const data = (await component.asyncData(asyncDataContext)) || {}
    options.data = () => ({ ...(returnedData || {}), ...data })
  }
  return mount(component, options)
}

```
notices how it merges `asyncData` with existing data in `options`

[...]

PR #250, opened 2021-09-14
feat: Expose the component instance

sarayourfriend:
Thanks for the awesome project! I've really enjoyed using it over at WordPress/Gutenberg and now introducing it to WordPress/openverse-frontend. It's been a great tool for learning and teaching people how to test with a focus on accessibility. Again, thanks so much!

I ran into this issue today so I decided to open a PR that would help me solve the issue, hope that's okay! This would also solve some of the issues found in this issue: #92

By exposing the `vm` from the `@vue/test-utils` wrapper, it'll allow doing things like calling `Component.fetch.call(wrapper.vm)` which is necessary becuase by default Vue doesn't call the "extra" Nuxt lifecycle methods (more about those here https://nuxtjs.org/docs/2.x/components-glossary/pages-fetch)

Maybe there's another more complex approach that would work more seamlessly (like adding a plugin that calls these methods for you or something) but this seems like the simplest option, just exposing the component instance that you'd get if you just used `@vue/test-utils` directly.

### DIFF
--- a/src/render.js
+++ b/src/render.js
@@ -75,6 +75,7 @@ function render(
     html: () => wrapper.html(),
     emitted: () => wrapper.emitted(),
     updateProps: _ => wrapper.setProps(_),
+    vm: wrapper.vm,
     ...getQueriesForElement(baseElement),
   }
 }
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -32,6 +32,7 @@ export interface RenderResult extends BoundFunctions<typeof queries> {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   emitted(): {[name: string]: any[][]}
   updateProps(props: object): Promise<void>
+  vm: Vue
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any


sarayourfriend:
@afontcu would it be possible for you to take a look at this PR. Is it possible that this feature could be merged? It would help fix some issues with using testing-library and Nuxt that we've run into at Openverse.

afontcu:
> @afontcu would it be possible for you to take a look at this PR. Is it possible that this feature could be merged? It would help fix some issues with using testing-library and Nuxt that we've run into at Openverse.

Hi! And sorry for the late reply, busy weeks and holidays and so on :) And thank you for your kind words! It's great to hear you folks at Wordpress are using it. How humbling! 😄 

I'm afraid this is unlikely to get merged as it kind of breaks two of the main principles for Vue Testing Lib:
1. Writing tests that resembled the way the software is used.
2. Have a symmetrical API with other testing-library flavours such as React – at least, as much as possible.

Sadly, by exposing `vm` we'd be kind of breaking both of them. I know (1) is not that clear here, as we're talking about setting up the right test env. However, exposing `vm` opens the door for a lot of arbitrary internal modifications of the component. That is, your use case is valid and hooking into Nuxt lifecycles might be sensible, but for instance one could also call private methods from `vm`

To solve your use case, maybe using the official Vue Test Utils library is a valid solution, as suggested in the linked issue? It exposes a myriad of options that would allow you to modify the `vm` while keeping Vue Testing Library closed to implementation details

That being said, if we find enough compelling cases for it we could reevaluate the situation and integrate some workaround.

sarayourfriend:
@afontcu Okay, thanks for the thoughtful response. I kind of figured that might be what the answer would be 🙂 I'll close this PR and we'll continue to use the regular Vue Test Utils directly in these cases.

Thanks again for all your hard work!

PR #276, opened 2022-07-16
fix: add missed optional prop for emmited function

thunderer199:
Vue test utils allow to pass event name  into `emitted()` to receive array of calls
`emitted().foo` can be written as `emitted('foo')`
You can see it [here](https://test-utils.vuejs.org/api/#emitted)


afontcu (approved):
nice catch, thanks!

PR #278, opened 2022-07-31
fix: align rerender with VTL and RTL

sand4rt:
fix for: https://github.com/testing-library/vue-testing-library/issues/198. To stir up this discussion.. I don't think you should deviate from RTL and STL unless there is a strong case against doing so. As far as i've seen it isn't there.


sand4rt:
It wasn't as easy as i thought. The component may not be destroyed and recreated..

PR #312, opened 2024-02-28
fix: check wrapper element to fix #311

### LINKED ISSUE #311: after manually calling the umount method, cleanupAtWrapper will report an error

After manually calling the umount method, cleanupAtWrapper will report an error TypeError: Cannot read properties of null (reading 'parentNode')

cleanup should not report an error

**Expected behavior**

**Screenshots**

image

**Related information:**

- `@testing-library/vue` version: 8.0.2
- `Vue` version: 3.2.33
- `node` version: 16
- `npm` (or `yarn`) version: pnpm v8

```javascript
import { createDragDropManager, DragDropManager } from 'dnd-core'
import { render, cleanup } from '@testing-library/vue'
import { useDndContextInjector } from '../DndContext'
import DndProvider from '../DndProvider'
import { TestBackend } from 'react-dnd-test-backend'
import { defineComponent, h } from 'vue-demi'
import { describe, afterEach, it, expect } from 'vitest'

describe('DndProvider', () => {
  afterEach(cleanup)

  it('stores DragDropManager in global context and cleans up on unmount', () => {
    let capturedManager

    const ChildComponent = defineComponent(() => {
      capturedManager = useDndContextInjector()
      return () => null
    })
    const TestProvider = defineComponent(() => {
      return () =>
        h(
          DndProvider,
          { backend: TestBackend },
          { default: () => h(ChildComponent) }
        )
    })

    const mountProvider = () => render(TestProvider)

    const globalInstance = (): DragDropManager =>
      (global as any)[Symbol.for('__VUE_DND_CONTEXT_INSTANCE__')] as any

    // Single mount & unmount works
    const root = mountProvider()
    expect(globalInstance()).toEqual(capturedManager)
    root.unmount()
    expect(globalInstance()).toEqual(null)

    // Two mounted components do a refcount
    const rootA = mountProvider()
    const rootB = mountProvider()
    expect(globalInstance()).toEqual(capturedManager)
    rootA.unmount()
    expect(globalInstance()).toEqual(capturedManager)
    rootB.unmount()
    expect(globalInstance()).toEqual(null)
  })
})


```

**Additional context**

hcg1023:
fix #311

### DIFF
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -89,31 +89,14 @@ test('unmounts', () => {
   expect(queryByTestId('node')).not.toBeInTheDocument()
 })
 
-test('use unmount before cleanup', () => {
-  const TestComponent = defineComponent((_, {slots}) => {
-    return () => slots.default?.()
-  })
+test('unmounts when no wrapper element is present', () => {
+  const Comp = defineComponent((_, ctx) => () => ctx.slots.default?.())
 
-  const {getByTestId, unmount, queryByTestId} = render({
-    render() {
-      return h(
-        TestComponent,
-        {},
-        {
-          default: () =>
-            h('div', {
-              'data-testid': 'node',
-            }),
-        },
-      )
-    },
+  const {unmount} = render({
+    render: () => h(Comp, () => h('div')),
   })
-
-  expect(getByTestId('node')).toBeInTheDocument()
 
   unmount()
-
-  expect(queryByTestId('node')).not.toBeInTheDocument()
 
   expect(() => cleanup()).not.toThrow()
 })
--- a/src/__tests__/render.js
+++ b/src/__tests__/render.js
@@ -90,24 +90,22 @@ test('unmounts', () => {
 })
 
 test('use unmount before cleanup', () => {
-  const ChildComponent = defineComponent(() => {
-    return () =>
-      h(
-        'div',
-        {
-          'data-testid': 'node',
-        },
-        ['Hi'],
-      )
-  })
-
-  const ParentComponent = defineComponent((_, {slots}) => {
+  const TestComponent = defineComponent((_, {slots}) => {
     return () => slots.default?.()
   })
 
   const {getByTestId, unmount, queryByTestId} = render({
     render() {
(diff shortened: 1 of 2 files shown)


afontcu:
hey! thank you for this :) it would be great if you could add a small test to assert that the bug is fixed – and to make sure it doesn't pop up again in the future. 

thank you!

hcg1023:
@afontcu I added test case

src/__tests__/render.js
```diff
@@ -87,3 +88,32 @@ test('unmounts', () => {
 
   expect(queryByTestId('node')).not.toBeInTheDocument()
 })
+
+test('use unmount before cleanup', () => {
```
afontcu, line 92:
```suggestion
test('unmounts when no wrapper element is present', () => {
```

src/render.js
```diff
@@ -61,6 +61,7 @@ function cleanup() {
 
 function cleanupAtWrapper(wrapper) {
   if (
+    wrapper.element &&
     wrapper.element.parentNode &&
     wrapper.element.parentNode.parentNode === document.body
```
afontcu, line 66: I guess this is quite old code, it could be clarified:

`wrapper.element?.parentNode?.parentNode === document.body`

src/__tests__/render.js
```diff
@@ -87,3 +88,32 @@ test('unmounts', () => {
...
+  const TestComponent = defineComponent((_, {slots}) => {
+    return () => slots.default?.()
+  })
+
+  const {getByTestId, unmount, queryByTestId} = render({
```
afontcu, line 97: I'd love to come up with the most succinct test that reproduces the issue. This would do it:

```js
test'unmounts when no wrapper element is present', () => {
  const Comp = defineComponent((_, ctx) => {
    return () => ctx.slots.default?.()
  })

  const {unmount} = render({
    render: () => h(Comp, () => h('div')),
  })

  unmount()
})
```

afontcu (changes requested):
Thanks for this! Added a few comments, almost there!

afontcu (approved):

PR #314, opened 2024-05-16
Add types for render function

cheyer:
Hi there 👋

I've been working with Vue3 / Storybook / Testing Library, and wondered why the props and slots typings work in Storybook. So I took inspiration in Storybook and added typings for the `render` function in vue testing library accordingly.

**What has been changed**
* added necessary deps
* added type for render function component props
* added type for render function component slots (`ComponentSlots` by `vue-component-type-helpers` by itself was not sufficient enough, since it does not allow slots of type string (e.g. `slots: { default: 'Test' }`), that's why we need the Typescript magic that Storybook uses)

**Reference:**
https://github.com/storybookjs/storybook/blob/ba69532715f162567cc17aa3a0de8ca918dfdd2c/code/renderers/vue3/src/public-types.ts#L63-L69


afontcu (approved):
This is great, thank you!

afontcu:
build is failing [due to some lint errors](https://github.com/testing-library/vue-testing-library/actions/runs/9120123600/job/25129354171?pr=314#step:5), would you mind taking a look?

cheyer:
> build is failing [due to some lint errors](https://github.com/testing-library/vue-testing-library/actions/runs/9120123600/job/25129354171?pr=314#step:5), would you mind taking a look?

Thanks for the review! The linting error should be fixed now!

PR #320, opened 2024-09-09
Fix #315 - RemoveIndexSignature deprecated in v4

### LINKED ISSUE #315: TypeScript v8.1.0: Cannot find module 'type-fest' or its corresponding type declarations - RemoveIndexSignature
TypeScript build of version 8.1.0 fails with:

> node_modules/@testing-library/vue/types/index.d.ts:10:36 - error TS2307: Cannot find module 'type-fest' or its corresponding type declarations.
> 
> import {RemoveIndexSignature} from 'type-fest
> 

This is due to `RemoveIndexSignature` being renamed to `OmitIndexSignature` in `type-fest` v3 and then being removed in a later version.

I would suggest specifying type-fest version 2.19 as a potential fix in the package.json. Version 3 and above appear to cause other issues.

Testing-library commit that causes the issue: https://github.com/testing-library/vue-testing-library/commit/d46ed8f757e77fe85b369847c21d394bb9706066#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R71

Deprecation notice of  `RemoveIndexSignature` from `type-fest`: https://github.com/sindresorhus/type-fest/issues/387

zymotik:
`RemoveIndexSignature` added in `type-fest` [v2.10.0](https://github.com/sindresorhus/type-fest/releases/tag/v2.10.0). Renamed to `OmitIndexSignature` in [v4.0.0](https://github.com/sindresorhus/type-fest/releases/tag/v4.0.0). `package.json` updated to reflect compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "lodash.merge": "^4.6.2",
     "msw": "^1.3.2",
     "tsd": "^0.29.0",
-    "type-fest": "~2.19",
+    "type-fest": ">=2.10 <4.0.0",
     "typescript": "^5.2.2",
     "vee-validate": "^4.11.8",
     "vue": "^3.3.5",


zymotik:
#315

G-Rath:
This won't fix the issue as you're only changing `devDependencies` which is not used by downstream packages - instead you want to add `type-fest` as an optional peer dependency dependency to express the range of versions that are supported if the dependency is in the tree

xiaoxiangmoe:
I think you should add type-fest to dependencies. Because if you add type-fest to peer dependency, it may conflict to other lib which using type-fest v4

